PR #4, opened 2026-03-23
security: restrict CORS from wildcard to validated same-origin

allenlinli:
## Problem

amux sets `Access-Control-Allow-Origin: *` on all responses. This is exploitable even behind a VPN or Tailscale:

1. User visits `evil.com` in their browser
2. `evil.com` JavaScript makes `fetch('http://amux-host:8822/api/file?path=~/.ssh/id_rsa')`
3. Browser sends the request because CORS `*` allows it
4. `evil.com` reads the response containing the SSH private key

**The user's browser acts as a proxy into the private network.** The attacker never needs direct access — they exploit the browser's position inside the network.

## Solution

Replace `Access-Control-Allow-Origin: *` with origin validation. Only these origins get a CORS header:

- `localhost` / `127.0.0.1` (local development)
- The server's LAN IP (same-network access)
- `*.ts.net` (Tailscale hostnames)

All other origins get no `Access-Control-Allow-Origin` header, causing the browser to block the response.

Also adds `Authorization` to `Access-Control-Allow-Headers` (needed for the auth token PR) and `X-Content-Type-Options: nosniff` + `X-Frame-Options: SAMEORIGIN` security headers.

## Test plan

- [ ] Dashboard at `localhost:8822` works (CORS header present)
- [ ] Dashboard at Tailscale hostname works (`*.ts.net` allowed)
- [ ] `curl -H 'Origin: https://evil.com' -v http://localhost:8822/api/sessions` — no `Access-Control-Allow-Origin` header in response
- [ ] `curl -H 'Origin: http://localhost:8822' -v http://localhost:8822/api/sessions` — `Access-Control-Allow-Origin: http://localhost:8822` in response
- [ ] 12 unit tests pass: `python3 tests/test_cors.py`

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_cors.py
+++ b/tests/test_cors.py
@@ -1,0 +1,106 @@
+"""Unit tests for CORS origin validation.
+
+These tests replicate the CORS origin validation logic from amux-server.py
+in isolation. The actual _cors() method sends HTTP headers — these tests
+verify the allow/deny decision logic.
+"""
+
+from urllib.parse import urlparse
+
+
+# ── Replicated CORS logic ────────────────────────────────────────────────────
+
+def _is_origin_allowed(origin: str, lan_ip: str = "192.168.1.100") -> bool:
+    """Standalone replica of the origin check in CCHandler._cors."""
+    if not origin:
+        return False
+    parsed = urlparse(origin)
+    host = parsed.hostname or ""
+    return (
+        host in ("localhost", "127.0.0.1", "0.0.0.0")
+        or host == lan_ip
+        or host.endswith(".ts.net")
+    )
+
+
+# ── Tests: allowed origins ───────────────────────────────────────────────────
+
+def test_localhost_allowed():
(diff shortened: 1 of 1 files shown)


esteininger (commented):
Good security improvement. The wildcard CORS origin is a real risk -- even behind Tailscale/VPN, a malicious page can use the browser as a proxy into the private network.

Your origin validation logic is solid (localhost, LAN IP, *.ts.net). Two notes:

1. The SSE endpoint fix (replacing hardcoded '*' with self._cors()) is an important catch that could easily be missed.
2. Adding 'Authorization' to allowed headers is forward-looking for the auth token PR.

This PR has merge conflicts with current main (the codebase has changed significantly since the base commit). I'm going to integrate this fix directly into the current codebase and credit you in the commit. Thanks for the contribution!

PR #6, opened 2026-03-23
security: replace weak path traversal checks with canonical resolution

allenlinli:
## Problem

The notes API uses `"..' in string` checks to prevent path traversal:

```python
if ".." in note_rel or note_rel.startswith("/"):
    return self._json({"error": "invalid path"}, 400)
note_path = CC_NOTES / note_rel
```

This is bypassable with:
- **URL encoding:** `%2e%2e/%2e%2e/etc/passwd` (`%2e` = `.`)
- **Double encoding:** `%252e%252e` decoded twice becomes `..`
- **Symlinks:** a symlink inside the notes directory pointing outside
- **Unicode normalization:** on some filesystems

An attacker can read, write, rename, and delete files **anywhere** the server process can access — not just inside the notes directory.

## Solution

Replace all `..` string checks with `_safe_note_path()`:

```python
def _safe_note_path(note_rel: str, base: Path = None) -> Path | None:
    candidate = (base / note_rel).resolve()
    try:
        candidate.relative_to(base.resolve())
    except ValueError:
        return None  # traversal detected
    return candidate
```

`Path.resolve()` canonicalizes the path (resolves `..`, symlinks, encoding) and `relative_to()` verifies the result is still inside the intended directory. This is the standard defense against path traversal in Python.

Applied to all notes endpoints:
- `GET /api/notes/:path` (read)
- `POST /api/notes/:path` (write)
- `PATCH /api/notes/:path` (rename/move)
- `POST /api/notes/trash/:name/restore` (restore)
- `DELETE /api/notes/trash/:name` (permanent delete)
- Share API note access

## Test plan

- [ ] Read/write/rename notes works normally
- [ ] `GET /api/notes/../../../etc/passwd` — 400 invalid path
- [ ] `GET /api/notes/%2e%2e/%2e%2e/etc/passwd` — 400 invalid path
- [ ] Trash restore with traversal — 400 invalid
- [ ] Move/rename to outside notes dir — 400 invalid
- [ ] 16 unit tests pass: `python3 tests/test_path_traversal.py`

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_path_traversal.py
+++ b/tests/test_path_traversal.py
@@ -1,0 +1,162 @@
+"""Unit tests for notes API path traversal protection.
+
+The old code used `".." in string` checks which are bypassable with
+URL encoding (%2e%2e), double encoding, or symlinks. The new
+_safe_note_path() uses Path.resolve() + relative_to() for canonical
+path validation.
+"""
+
+import tempfile
+from pathlib import Path
+
+
+# ── Replicated path traversal logic ──────────────────────────────────────────
+
+def _safe_note_path(note_rel: str, base: Path = None) -> Path | None:
+    """Standalone replica of _safe_note_path from amux-server.py."""
+    if not note_rel or note_rel.startswith("/"):
+        return None
+    candidate = (base / note_rel).resolve()
+    try:
+        candidate.relative_to(base.resolve())
+    except ValueError:
+        return None  # traversal detected
+    return candidate
+
+
+# ── Tests: valid note paths ──────────────────────────────────────────────────
+
+def test_simple_note():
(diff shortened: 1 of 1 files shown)


esteininger (approved):
Great catch. The "..' in string" check is a classic weak pattern bypassable with URL encoding, symlinks, and Unicode normalization. Your _safe_note_path() using Path.resolve() + relative_to() is the correct canonical defense. The implementation is clean -- handles both notes and trash directories, correctly blocks absolute paths, and tests cover symlink bypass which many contributors miss. Merging this. Thanks Allen\!

PR #18, opened 2026-04-10
feat(ui): default model settings + 1M context options

kmichels:
## Summary
- Adds a **Default Model** section to the Settings UI — dropdown to pick the model applied to new sessions, persisted to `defaults.env` via `PATCH /api/settings/default-model`
- Adds `claude-opus-4-6[1m]` and `claude-sonnet-4-6[1m]` to both the per-session model picker and the settings dropdown (1M extended context variants)
- Makes the "Default" label in the per-session model picker dynamic — shows the actual configured default instead of hardcoded "sonnet"

## How it works
- `_get_default_model()` reads `CC_DEFAULT_FLAGS` from `~/.amux/defaults.env` and extracts the `--model` value
- Server injects `window._AMUX_DEFAULT_MODEL` into the dashboard HTML at render time
- Settings UI reads/writes via `GET/PATCH /api/settings/default-model`
- Changing the default updates the per-session dropdown label in real time

## Test plan
- [ ] Open Settings, verify Default Model dropdown shows current default
- [ ] Change default model, verify toast confirms and per-session "Default (X)" label updates
- [ ] Start a new session with no explicit model — verify it uses the configured default
- [ ] Verify `~/.amux/defaults.env` is written correctly after save

### DIFF
--- a/amux-server.py
+++ b/amux-server.py
@@ -4653,6 +4653,19 @@ def _ensure_memory(name: str, work_dir: str):
     _write_claude_memory(name, work_dir)
 
 
+def _get_default_model() -> str:
+    """Return the default model from defaults.env, or 'sonnet' as fallback."""
+    defaults_file = CC_HOME / "defaults.env"
+    if defaults_file.exists():
+        dcfg = parse_env_file(defaults_file)
+        flags = dcfg.get("CC_DEFAULT_FLAGS", "")
+        import re
+        m = re.search(r"--model\s+(\S+)", flags)
+        if m:
+            return m.group(1)
+    return "sonnet"
+
+
 def start_session(name: str, extra_flags: str = "", _skip_conv_id: bool = False) -> tuple[bool, str]:
     """Start a session headless (no attach). Returns (success, message)."""
     f = CC_SESSIONS / f"{name}.env"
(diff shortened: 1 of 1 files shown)


kmichels:
Withdrawing this PR — the `claude-opus-4-6[1m]` and `claude-sonnet-4-6[1m]` options ship a latent shell-injection bug.

`PATCH /api/settings/default-model` writes the model value verbatim into `defaults.env` as `CC_DEFAULT_FLAGS="--model claude-opus-4-6[1m]"`. In `start_session()` that string gets concatenated into a shell command and handed to `tmux new-session`, which runs it via the user's login shell. zsh (default on macOS) interprets `[1m]` as a glob character class, finds no match, and aborts the spawn with `zsh:1: no matches found: claude-opus-4-6[1m]`. Every new session start is bricked until `defaults.env` is hand-edited.

Same issue affects the per-session model picker (`PATCH /api/sessions/{name}` with `model:` body) which writes into `CC_FLAGS` the same way, and the bash CLI's `cmd_start` which has its own identical concat.

Fix in flight: `_shell_quote_flags()` helper at the spawn boundary in both `amux-server.py` and the `amux` bash CLI, plus regex validation at the PATCH endpoints. Will resubmit as a new PR once the fix lands and is tested locally.

PR #19, opened 2026-04-11
fix(spawn): shell-safe flag quoting + default-model settings (replaces #18)

kmichels:
## Summary

Replaces closed #18 with a corrected version. #18 added a Default Model dropdown with 1M-context options like `claude-opus-4-6[1m]`, but selecting one bricked every new session start with `zsh:1: no matches found: claude-opus-4-6[1m]`. zsh interprets `[1m]` as a glob character class and aborts the spawn command before claude is invoked.

The root cause was broader than #18: stored flag strings were concatenated raw into shell commands in both spawn paths (`amux-server.py:start_session` and `amux:cmd_start`). The 1M model IDs just made an existing latent bug reachable from the UI. This PR fixes it at the consumer boundary so any producer (settings UI, REST API, bash CLI, hand-edited `.env` files) is automatically safe, AND includes the corrected version of the #18 default-model UI so users can actually use the 1M context variants.

## What's in this PR

**The fix from #18 (default-model UI):**
- Settings → Default Model dropdown
- `claude-opus-4-6[1m]` and `claude-sonnet-4-6[1m]` in the per-session model picker
- `PATCH /api/settings/default-model` endpoint
- Dynamic "Default (X)" label reflecting the configured default

**Plus the shell-safe quoting fix (the new part):**
- `_shell_quote_flags()` in `amux-server.py`: tokenize stored flag strings via `shlex.split` and re-quote each token via `shlex.quote`. Applied at all 4 concat sites in `start_session()`. On malformed input, wraps the whole raw string as a single `shlex.quote`'d literal so the security invariant holds.
- `quote_flags()` + `quote_argv()` in `amux` (bash CLI): delegate to python3 shlex via a shared `_run_py_safe` runner. Uses python3 shlex instead of bash's `printf '%q'` because `printf '%q'` produces bash-only `$'...'` syntax that breaks under `dash`. shlex.quote is universal POSIX single-quoting.
- `cmd_start` inline args (`amux start NAME -- …`) now use `quote_argv` instead of the broken `$*` form. Previously `$*` flattened argv back into a single string, losing quoting boundaries when the nested shell re-evaluated it — this enabled both quote-loss (`--system-prompt "hello world"` became 3 args) and command injection (`--system-prompt "wait; echo BOOM"` actually ran `echo BOOM`).
- Source-guard at bottom of `amux` lets the test suite `source` it without triggering CLI dispatch.

**Defense in depth:**
[...]

PR #20, opened 2026-04-16
feat(ui): add Claude Opus 4.7 model options

kmichels:
Anthropic released Claude Opus 4.7 today ([announcement](https://www.anthropic.com/news/claude-opus-4-7), [docs](https://docs.claude.com/en/docs/about-claude/models/overview)). Adds it to both model selectors:

- Settings → Default Model dropdown
- Per-session model picker in the edit overlay

Per the docs, Opus 4.7 has **1M tokens of context natively** (no beta required). Included both `claude-opus-4-7` and `claude-opus-4-7[1m]` to stay consistent with the existing 4.6 pattern — the `[1m]` variant is a no-op API-side but exists in case a future harness uses the suffix as an explicit opt-in toggle. Easy to drop the `[1m]` row if you'd rather not carry both.

Order in the dropdown: new 4.7 entries go above the 4.6 entries so the latest model is at the top of each family section.

## Notes

- Shell-quoting for the `[1m]` glob-looking suffix is already handled by `_shell_quote_flags()` from #19 (`f9b7b60`) — no new quoting work needed.
- `claude-haiku-4-5-20251001` is still the current Haiku (`claude-haiku-4-5` alias), so no Haiku changes.

## Test plan

- [x] `python3 -c "import ast; ast.parse(open('amux-server.py').read())"` passes
- [x] Server auto-restart on file save picks up changes
- [x] Both dropdowns render the new options (verified against `curl $AMUX_URL/`)
- [ ] Spawn a session with `claude-opus-4-7[1m]` and confirm 1M context window is honored (can only confirm locally)


esteininger:
Thanks for the PR! These options were already added to main independently — the dropdown now has both claude-opus-4-7 and the [1M] variant. Closing as already implemented.

PR #21, opened 2026-04-17
feat(ui): alphabetical sort toggle for session list

kmichels:
## Problem

Session cards re-sort on every activity change because the default order is `pinned → running → status → last_activity`. When you're hunting for a session to click, another session going active can shift the card you're aiming at out from under your cursor. This is especially annoying in the group layout where sessions jump between buckets as their status transitions.

## Change

Adds a third sort-order control next to the existing layout buttons (list / group / grid): an **A↓Z** toggle that switches between the existing natural sort and a stable alphabetical one.

- **Natural** (default, unchanged): `pinned → running → status → last_activity`
- **Alpha** (new): `pinned → name` — pinned sessions still pin to top, everything else in stable name order

Preference persists in `localStorage` under `amux_sort_mode`. Works in all three layout modes: in group view, each status bucket sorts alphabetically; in list and grid views, sessions are one flat alpha list.

## Behavior notes

- Pinned-first is preserved under alpha — pinning still has meaning as "keep this session visible regardless of sort"
- Drag-to-reorder via `Sortable` is disabled while alpha is active (it would re-sort on next render anyway). Flip back to natural to drag — existing `cardOrder` is untouched, so manual ordering is preserved across sort-mode toggles
- Search/tag filters take precedence over sort-mode in existing paths; alpha kicks in when no filter is active
- Server-side sort in `list_sessions()` is left alone — every client render path already re-sorts locally, so no API change is needed

## Test plan

- [x] Python syntax check (`ast.parse`)
- [x] Dashboard renders the A↓Z button in `tile-controls`
- [x] Toggling the button persists across reload via localStorage
- [x] Alpha mode sorts correctly in list, group (per-bucket), and grid layouts
- [x] Pinned sessions still pin to top under alpha
- [x] Drag handles no-op under alpha and re-activate on flip back

## What this doesn't do

- No server-side sort change — kept scope to the client
- Doesn't reorder the status *buckets* in group mode, only the cards within each bucket. The bucket order (Working → Needs Input → Idle → Stopped) is still semantic
- No new localStorage migration — fresh browsers get the natural default

### DIFF
--- a/amux-server.py
+++ b/amux-server.py
@@ -9954,6 +9954,7 @@ DASHBOARD_HTML = r"""<!DOCTYPE html>
     <button class="tile-btn" id="tile-list-btn" onclick="setLayoutMode('list')" title="List view">&#x2630;</button>
     <button class="tile-btn" id="tile-group-btn" onclick="setLayoutMode('group')" title="Group by status" style="font-size:0.75rem;font-weight:700;">#</button>
     <button class="tile-btn tile-grid-only" id="tile-grid-btn" onclick="setLayoutMode('grid')" title="Grid view">&#x268F;</button>
+    <button class="tile-btn" id="tile-sort-btn" onclick="toggleSortMode()" title="Sort alphabetically (pinned stay on top, order stops shifting)" style="font-size:0.7rem;font-weight:700;letter-spacing:-0.5px;">A&#x2193;Z</button>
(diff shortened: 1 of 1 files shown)


esteininger:
Applied in 744e21a — thank you! Clean implementation, the stable alpha sort is a nice addition. Kept the design as-is: localStorage persistence, pinned-first preserved, Sortable disabled during alpha mode.

PR #22, opened 2026-04-17
fix(ui): stop accidental Files-tab teleport from session card clicks

kmichels:
## The bug

Clicking a session card in the Sessions tab would sometimes route you into the Files tab for that session instead of expanding the card. Then clicking the Sessions tab to return would dump you into the Terminal/peek view rather than back to the session list. Felt intermittent, but had a deterministic cause.

## Root cause

Two coupled click handlers:

1. **`_renderSessionCard` → `.card-dir-path`** (the directory line under the card header) had an `onclick` that called `openExplore(...)` → `switchView('files')`. The dir path is visually prominent inside the card but narrow vertically; when cards shift mid-click due to natural-sort reorder on activity changes, the cursor frequently landed on a neighbor's path → routed to Files.

2. **`switchView('sessions')`** had a deferred action: if `_exploreSession` was set (from step 1), it scheduled `setTimeout(() => openPeek(_sess), 50)`. So clicking the Sessions tab to recover would briefly show Sessions, then get teleported into the peek view.

## Changes

- Remove the `onclick` from `.card-dir-path`. The path is now informational text with a tooltip showing the full directory.
- Add **Browse files** as a card-menu item (`⋯` menu) next to **Peek terminal**. Explicit affordance, no accidental activation.
- In `switchView('sessions')`, clear `_exploreSession` without the auto-peek side effect. Sessions tab now reliably shows Sessions.

## What this doesn't break

- `_exploreSession` is still set by `openExplore()` and cleared only on Sessions-tab return, so the **Set dir for <session>** button inside Files view (uses `_exploreSession`) is unaffected.
- Card-menu pattern already has nine action items; adding a tenth fits the existing design.
- Click-to-Peek via the `card-preview-lines` element (line ~12566) is untouched — double-tap on header still opens peek too.

## Test plan

- [x] Python syntax (`ast.parse`)
- [x] Clicking anywhere on card body now reliably toggles expand/collapse
- [x] Card-menu &rarr; Browse files opens Files view rooted at the session's dir
- [x] Returning to Sessions tab from Files shows the session list, no peek teleport
- [x] Set-dir-for-session button in Files view still works

### DIFF
--- a/amux-server.py
+++ b/amux-server.py
@@ -12504,6 +12504,7 @@ function render() {
           <button class="card-menu-btn" onclick="event.stopPropagation();toggleMenu('${s.name}')" title="Options">&#x22EF;</button>
           <div class="card-menu" id="menu-${s.name}">
           <div class="card-menu-item" onclick="event.stopPropagation();closeAllMenus();openPeek('${s.name}')"><span class="mi">&#x1F4BB;</span> Peek terminal</div>
+          ${s.dir ? `<div class="card-menu-item" onclick="event.stopPropagation();closeAllMenus();openExplore('${s.dir.replace(/'/g,"\\'")}','${s.name.replace(/'/g,"\\'")}')"><span class="mi">&#x1F4C1;</span> Browse files</div>` : ''}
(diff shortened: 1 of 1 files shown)


esteininger:
Applied in 744e21a — good catch on the teleport bug. The dir-path onclick was definitely causing confusion. Browse files is now in the ... menu where it belongs.

PR #25, opened 2026-04-23
feat: graceful session restart with name-based resume

kmichels:
## Summary

- Fixes session context loss on restart when user has `/clear`ed during a session
- `stop_session` now sends `/exit` gracefully instead of hard-killing tmux — mosh stays connected, scrollback preserved
- Sessions resume by name (`claude --resume "<name>"`) instead of UUID, which survives `/clear` forks
- tmux sessions persist across restarts — only the Claude process cycles

## Root cause

When a user types `/clear` in Claude Code, Claude forks to a new session-id and JSONL file. amux's stored `cc_conversation_id` was never updated, so `--resume <stale_uuid>` on restart loaded the pre-`/clear` state — silently discarding hours or days of conversation.

Confirmed on `college-advisor-1`: user `/clear`ed on Apr 21, but restart on Apr 23 resurrected Apr 17 state.

## What changed

- **`stop_session`**: reads session name from Claude's PID file (or sets one via `/rename`), sends `/exit`, polls for shell prompt (15s timeout, hard-kill fallback)
- **`start_session`**: reuses existing tmux session, launches Claude via `send-keys` into a persistent shell. Resume strategy: `cc_session_name` > `cc_conversation_id` (migration) > fresh `--name`
- **`is_running`**: checks Claude state (not just tmux existence)
- **Auto-restart paths**: error-recovery uses `_hard_kill_claude` (kills Claude process only, keeps tmux)
- **API**: `POST /stop` returns 202, runs in bounded ThreadPoolExecutor
- **JS**: `doRestart` poll timeout extended from 5s to 20s
- Per-session `threading.RLock` prevents stop/start races
- Session name validation (`^[a-zA-Z0-9][a-zA-Z0-9_.\-]*$`) before shell use

## Migration

Existing UUID-based sessions auto-migrate on first stop+start cycle. No manual intervention needed (except `college-advisor-1` which needs a manual `/rename` in the live session).

## Test plan

- [ ] Restart session without prior `/clear` — resumes correctly
- [ ] Restart session after `/clear` — resumes post-clear state
- [ ] Custom-named session survives restart
- [ ] Unnamed session gets auto-named before `/exit`
- [ ] Hard-kill fallback if Claude hangs on `/exit`
- [ ] Multiple sessions sharing work_dir restart independently
- [ ] Mosh connection survives restart
- [ ] First-ever start of new session works
- [ ] `--resume` failure falls back to fresh `--name` start


esteininger:
Merged manually with conflict resolution. Great work on the graceful restart fix — the /clear context loss bug was a real pain point. Fixed the hardcoded mosh command (konrad@bot-mini → dynamic hostname) to keep the single-codebase rule. Thanks for the contribution!

PR #27, opened 2026-04-28
fix: dashboard restart never relaunched Claude after stop

kmichels:
## Summary

The dashboard's "Restart" button (running session card → Restart, card menu → Restart) leaves the tmux pane at a bare shell prompt — Claude is killed by /stop but never relaunched. User has to manually run `claude` in the pane to recover.

## Root cause

Two interacting bugs in two layers, both required to reproduce:

### Layer 1: dashboard polled the wrong endpoint

Three commits combined into a deadlock:
1. `51c6f7c` (Feb 22) — `/api/sessions` list endpoint changed `running` from `is_running(name)` to `tmux_name(name) in tmux_info` (perf optimization, 100×).
2. `bb1f935` (Apr 8) — `doRestart` added; polled `fetchSessions()` for `s.running === false` then called `doStart`.
3. `stop_session` design — keeps tmux alive (only kills Claude, not the pane).

Result: after /stop, the list endpoint's `running` stays `true` forever (tmux is alive). Polling timed out at 20s, fired `showToast("Stop didn't take effect, try again")`, and skipped `doStart` entirely.

### Layer 2: even the "right" endpoint was lying

The single-session endpoint `/api/sessions/{name}/info` uses `is_running()` which calls `_claude_ui_visible()` to detect "Claude UI present in last 12 lines of pane". That check matched ANY line starting with a dingbat character (`✀`–`➿`). The Claude prompt char `❯` (U+276F) is in that range — so the user's `❯ /exit` input sits in scrollback within the 12-line window and trips the heuristic forever after Claude exits.

This second bug also affected `stop_session`: it uses the same `_at_shell_prompt` detector to know when Claude has exited, so every restart was hitting the 15s graceful-exit timeout and falling through to hard-kill, even when Claude exited cleanly in <1s.

## Empirical confirmation

```
2026-04-28 01:52:48 [phone] POST /api/sessions/scripts-4/stop 202 8ms
(no matching POST /start anywhere)
~01:54   user manually ran 'claude --dangerously-skip-permissions' to recover
```

Live verification on `coaching` session:
- Before fix: `/info` reported `running:true` for 30s+ after /stop, polling timed out, /start short-circuited with "already running" because `is_running` was lying. Claude actually died and was never relaunched.
- After fix: `/info` flipped to `false` at T+1.2s after /stop. Polling exited cleanly, /start fired and relaunched Claude.

Headless WebKit run via Playwright (`window.doRestart('coaching')` called directly):
- T+1.07s: `running=false` observed
- T+5.28s: `running=true` again — full restart cycle in 5.3s
- 0 page errors / 0 console errors / 0 console warnings / 0 request failures

## Fix

**JS (commits 1–5):** switch `doRestart` poll target from `fetchSessions()` (list) to `/api/sessions/{name}/info` (single). Plus hardening from successive panel review rounds:
- Pre-check for already-stopped → skip /stop entirely (avoids TOCTOU where `_bg_stop` kills a freshly-launched Claude)
- Bare `fetch` with explicit 404 handling (apiCall returns null on 4xx and toasts a generic error)
- Module-level Set guards against overlapping calls from card menu + peek menu
- `AbortController + setTimeout` for hard fetch timeouts (Safari 16 compat over `AbortSignal.timeout()`)
- Wall-clock deadline instead of fixed iteration count (so 3s per-poll timeout doesn't extend the 30s window to 3 min)
- Polling window 30s with 1s warmup, since worst-case `stop_session` is ~21s

**Python (commit 6):** exclude `❯` from `_claude_ui_visible`'s dingbat-prefix scan. Real spinner chars are `✻ ✶ ⏺ ✦` etc. — never `❯` — so this removes a false positive without breaking genuine spinner detection.

## Out of scope

- Semantic inconsistency between list endpoint (line 4576: tmux-existence) and single endpoint (line 4609: process-state). Worth a follow-up to either fix the list endpoint with a TTL cache on `is_running`, or rename the keys. Not bundled here.
- Bash CLI `amux restart` (line ~2494). Already works via RLock serialization in `_get_session_lock` — no change needed, even with the `_claude_ui_visible` bug, because RLock blocks /start until _bg_stop completes.
- Auto-restart watcher (line 1622-1636). Calls `start_session` directly. The `_claude_ui_visible` fix actually helps it too — previously the watcher couldn't detect Claude exits cleanly on sessions where `❯ <last input>` was still in scrollback.

## Test plan

- [x] **Server-side `is_running()` correctly flips false ~1s after `/exit`** — verified with curl on `coaching`, `/info` returns `running:false` at T+1.2s
- [x] **Full restart cycle via API** — /stop → /info flips false → /start succeeds → Claude relaunched, `start_count` increments (14 → 15 → 16 → 17 across three runs)
- [x] **Dashboard JS in headless WebKit** — Playwright loaded the dashboard, called `window.doRestart('coaching')`, observed full cycle in 5.3s with 0 page/console errors. Verifies the fixed JS parses and runs end-to-end in the browser engine the PWA actually targets.
- [ ] Manual: dashboard Restart on session whose Claude already crashed → pre-check fires, skips /stop, doStart launches Claude
- [ ] Manual: dashboard Restart, immediately click again → second click toasts "Restart already in progress"
- [ ] Manual: dashboard Restart on non-existent session → "Session no longer exists" toast within 1s

### DIFF
--- a/amux-server.py
+++ b/amux-server.py
@@ -14134,7 +14134,7 @@ async function doRestart(name) {
     let alreadyStopped = false;
     let sessionGone = false;
     try {
-      const pre = await fetch(INFO_URL);
+      const pre = await fetch(INFO_URL, { signal: AbortSignal.timeout(5000) });
       if (pre.status === 404) {
         sessionGone = true;
       } else if (pre.ok) {
@@ -14168,7 +14168,7 @@ async function doRestart(name) {
     const MAX_POLLS = 58; // 1s warmup + 58 * 500ms ≈ 30s total
     for (let i = 0; i < MAX_POLLS; i++) {
       try {
-        const r = await fetch(INFO_URL);
+        const r = await fetch(INFO_URL, { signal: AbortSignal.timeout(3000) });
         if (r.status === 404) {
           showToast('Session no longer exists');
           return;
--- a/amux-server.py
+++ b/amux-server.py
@@ -14111,24 +14111,79 @@ async function copyMoshCmd(name) {
   }
 }
 
+// Module-level guard so multiple Restart entry points (card menu, peek menu)
+// can't pile up overlapping polling loops + queued /start commands.
+const _restartingSessions = new Set();
+
 async function doRestart(name) {
-  await apiCall(API + '/api/sessions/' + name + '/stop', { method: 'POST' });
-  // Poll until stopped (up to 20s for graceful /exit). If it never stops,
-  // bail out instead of racing doStart against a still-running session.
-  let stopped = false;
-  for (let i = 0; i < 40; i++) {
-    await new Promise(r => setTimeout(r, 500));
-    await fetchSessions();
-    if (!sessions.find(s => s.name === name && s.running)) {
-      stopped = true;
-      break;
-    }
-  }
-  if (!stopped) {
-    showToast("Stop didn't take effect, try again");
+  if (_restartingSessions.has(name)) {
+    showToast('Restart already in progress');
     return;
   }
-  await doStart(name);
+  _restartingSessions.add(name);
+  try {
+    const encodedName = encodeURIComponent(name);
+    const SINGLE_URL = API + '/api/sessions/' + encodedName;
+    const STOP_URL = SINGLE_URL + '/stop';
+
+    // Pre-check: if Claude already isn't running (crashed, externally killed),
+    // skip /stop entirely and go straight to doStart. Without this, the queued
+    // _bg_stop worker can race a freshly-launched Claude and kill it.
+    // Use raw fetch so we can distinguish 404 from 5xx; apiCall flattens both.
+    try {
+      const pre = await fetch(SINGLE_URL);
+      if (pre.status === 404) {
+        showToast('Session no longer exists');
+        return;
+      }
+      if (pre.ok) {
+        const data = await pre.json();
+        if (!data.running) {
+          await doStart(name);
+          return;
+        }
+      }
+    } catch (e) {
+      // Network blip on pre-check — fall through to standard stop+poll flow.
+    }
+
+    const stopResp = await apiCall(STOP_URL, { method: 'POST' });
+    if (!stopResp) return; // apiCall already surfaced the error
+
+    // Poll the SINGLE-session endpoint, not the list. The list endpoint reports
+    // running == "tmux session exists", which never flips false because
+    // stop_session keeps tmux alive. The single endpoint uses is_running()
+    // which correctly returns false once Claude exits.
+    // Worst case server-side stop is ~21s (rename + 15s exit + 1s settle), so
+    // poll for 30s with a 1s warmup before the first poll.
+    await new Promise(r => setTimeout(r, 1000));
+    let stopped = false;
+    const MAX_POLLS = 58; // 1s warmup + 58 * 500ms ≈ 30s total
+    for (let i = 0; i < MAX_POLLS; i++) {
+      try {
+        const r = await fetch(SINGLE_URL);
+        if (r.status === 404) {
+          showToast('Session no longer exists');
+          return;
+        }
+        if (r.ok) {
+          const data = await r.json();
+          if (!data.running) { stopped = true; break; }
+        }
+      } catch (e) {
+        // Network blip — keep polling.
+      }
+      await new Promise(r => setTimeout(r, 500));
+    }
+    if (!stopped) {
+      showToast("Stop didn't take effect, try again");
+      return;
+    }
+    await fetchSessions(); // refresh card UI before doStart
+    await doStart(name);
+  } finally {
+    _restartingSessions.delete(name);
+  }
 }
 
 // ── Sending indicator ──
--- a/amux-server.py
+++ b/amux-server.py
@@ -14123,37 +14123,44 @@ async function doRestart(name) {
   _restartingSessions.add(name);
   try {
     const encodedName = encodeURIComponent(name);
-    const SINGLE_URL = API + '/api/sessions/' + encodedName;
-    const STOP_URL = SINGLE_URL + '/stop';
+    const INFO_URL = API + '/api/sessions/' + encodedName + '/info';
+    const STOP_URL = API + '/api/sessions/' + encodedName + '/stop';
 
     // Pre-check: if Claude already isn't running (crashed, externally killed),
-    // skip /stop entirely and go straight to doStart. Without this, the queued
-    // _bg_stop worker can race a freshly-launched Claude and kill it.
+    // skip /stop entirely. Without this, the queued _bg_stop worker can race a
+    // freshly-launched Claude and kill it.
     // Use raw fetch so we can distinguish 404 from 5xx; apiCall flattens both.
+    // (window.fetch is monkey-patched at line 11878 to inject auth headers.)
+    let alreadyStopped = false;
+    let sessionGone = false;
     try {
-      const pre = await fetch(SINGLE_URL);
+      const pre = await fetch(INFO_URL);
       if (pre.status === 404) {
-        showToast('Session no longer exists');
-        return;
-      }
-      if (pre.ok) {
+        sessionGone = true;
+      } else if (pre.ok) {
         const data = await pre.json();
-        if (!data.running) {
-          await doStart(name);
-          return;
-        }
+        if (!data.running) alreadyStopped = true;
       }
     } catch (e) {
       // Network blip on pre-check — fall through to standard stop+poll flow.
+    }
+    if (sessionGone) {
+      showToast('Session no longer exists');
+      return;
+    }
+    if (alreadyStopped) {
+      await fetchSessions();
+      await doStart(name);
+      return;
     }
 
     const stopResp = await apiCall(STOP_URL, { method: 'POST' });
     if (!stopResp) return; // apiCall already surfaced the error
 
-    // Poll the SINGLE-session endpoint, not the list. The list endpoint reports
-    // running == "tmux session exists", which never flips false because
-    // stop_session keeps tmux alive. The single endpoint uses is_running()
-    // which correctly returns false once Claude exits.
+    // Poll the SINGLE-session info endpoint, not the list. The list endpoint
+    // reports running == "tmux session exists", which never flips false because
+    // stop_session keeps tmux alive. /info uses is_running() which correctly
+    // returns false once Claude exits.
     // Worst case server-side stop is ~21s (rename + 15s exit + 1s settle), so
     // poll for 30s with a 1s warmup before the first poll.
     await new Promise(r => setTimeout(r, 1000));
@@ -14161,7 +14168,7 @@ async function doRestart(name) {
     const MAX_POLLS = 58; // 1s warmup + 58 * 500ms ≈ 30s total
     for (let i = 0; i < MAX_POLLS; i++) {
       try {
-        const r = await fetch(SINGLE_URL);
+        const r = await fetch(INFO_URL);
         if (r.status === 404) {
           showToast('Session no longer exists');
           return;
--- a/amux-server.py
+++ b/amux-server.py
@@ -1498,10 +1498,14 @@ def _claude_ui_visible(clean_output: str) -> bool:
         ls = l.strip().lower()
         if "\u23f5\u23f5" in l or "bypass permissions" in ls or "plan mode" in ls:
             return True
-    # Check last 12 lines for an active spinner (dingbat + ellipsis)
+    # Check last 12 lines for an active spinner (dingbat-prefixed status line
+    # like "\u273b Crunched for 1m 38s"). Exclude U+276F \u276f \u2014 that's Claude's input
+    # prompt, also the user's typed input in scrollback. After /exit, lines
+    # like "\u276f /exit" stay in scrollback within this 12-line window and were
+    # tripping a false positive that made is_running() return True forever.
     for l in lines[-12:]:
         s = l.strip()
-        if s and "\u2700" <= s[0] <= "\u27bf":
+        if s and "\u2700" <= s[0] <= "\u27bf" and s[0] != "\u276f":
             return True
     return False
 
--- a/amux-server.py
+++ b/amux-server.py
@@ -14173,11 +14173,14 @@ async function doRestart(name) {
     // stop_session keeps tmux alive. /info uses is_running() which correctly
     // returns false once Claude exits.
     // Worst case server-side stop is ~21s (rename + 15s exit + 1s settle), so
-    // poll for 30s with a 1s warmup before the first poll.
+    // poll for 30s with a 1s warmup before the first poll. Use a wall-clock
+    // deadline rather than an iteration count: with a 3s per-poll timeout,
+    // a fixed 60-iter loop would actually run for >3 minutes if the network
+    // is hanging.
     await new Promise(r => setTimeout(r, 1000));
     let stopped = false;
-    const MAX_POLLS = 58; // 1s warmup + 58 * 500ms ≈ 30s total
-    for (let i = 0; i < MAX_POLLS; i++) {
+    const deadline = Date.now() + 30000;
+    while (Date.now() < deadline) {
       try {
         const r = await _fetchTimeout(INFO_URL, 3000);
         if (r.status === 404) {
@@ -14191,6 +14194,7 @@ async function doRestart(name) {
       } catch (e) {
         // Network blip — keep polling.
       }
+      if (Date.now() >= deadline) break;
       await new Promise(r => setTimeout(r, 500));
     }
     if (!stopped) {
--- a/amux-server.py
+++ b/amux-server.py
@@ -14115,6 +14115,17 @@ async function copyMoshCmd(name) {
 // can't pile up overlapping polling loops + queued /start commands.
 const _restartingSessions = new Set();
 
+// fetch() with a hard timeout, so a hanging network request can't trap the
+// caller inside an await. Using AbortController + setTimeout instead of
+// AbortSignal.timeout() for broader Safari/iOS compat (AbortSignal.timeout
+// is Safari 16+; AbortController is universal). Aborts surface as a
+// DOMException, which the caller's try/catch handles like any network error.
+function _fetchTimeout(url, ms) {
+  const ctl = new AbortController();
+  const t = setTimeout(() => ctl.abort(), ms);
+  return fetch(url, { signal: ctl.signal }).finally(() => clearTimeout(t));
+}
+
 async function doRestart(name) {
   if (_restartingSessions.has(name)) {
     showToast('Restart already in progress');
@@ -14134,7 +14145,7 @@ async function doRestart(name) {
     let alreadyStopped = false;
     let sessionGone = false;
     try {
-      const pre = await fetch(INFO_URL, { signal: AbortSignal.timeout(5000) });
+      const pre = await _fetchTimeout(INFO_URL, 5000);
       if (pre.status === 404) {
         sessionGone = true;
       } else if (pre.ok) {
@@ -14168,7 +14179,7 @@ async function doRestart(name) {
     const MAX_POLLS = 58; // 1s warmup + 58 * 500ms ≈ 30s total
     for (let i = 0; i < MAX_POLLS; i++) {
       try {
-        const r = await fetch(INFO_URL, { signal: AbortSignal.timeout(3000) });
+        const r = await _fetchTimeout(INFO_URL, 3000);
         if (r.status === 404) {
           showToast('Session no longer exists');
           return;


kmichels:
# Review Panel: PR #27: fix(ui): doRestart never called doStart due to broken running flag

**Reviewers:** Security, Architecture, Correctness
**Date:** 2026-04-28

## Moderated Findings

Here is a consolidated and prioritized list of findings from the reviews.

### Prioritized Findings List

**Blocker**

*   **Missing Authentication in Raw `fetch` Calls**
    *   The new raw `fetch` calls for pre-checking and polling likely omit authentication headers (e.g., `Authorization: Bearer <token>`) that are normally handled by the `apiCall` wrapper.
    *   This will cause the API to return `401 Unauthorized`, breaking the restart functionality and causing the polling loop to always time out.
    *   **Flagged by:** Security, Correctness

**High**

*   **Backend API Contract Inconsistency**
    *   The backend API has a severe design flaw: the `/api/sessions` list endpoint and the `/api/sessions/{name}` detail endpoint return different semantic meanings for the `running` field.
    *   This forces the client to know which endpoint to poll for the "true" state, breaking RESTful principles and making the API untrustworthy.
    *   **Flagged by:** Architecture
*   **Backend `/stop` Endpoint Is Not Idempotent**
    *   The backend has a race condition where calling `/stop` on an already-stopped session queues a background worker that can kill a newly started process.
    *   The client-side pre-check is a workaround for a fundamental backend state management flaw. The `/stop` endpoint should be idempotent.
    *   **Flagged by:** Architecture
*   **`doStart` Errors are Swallowed by `try/catch` Block**
    *   The `await doStart(name)` call is incorrectly placed inside a `try/catch` block intended for network errors during the pre-check `fetch`.
    *   If `doStart` fails, its error will be silently caught, and the code will incorrectly proceed to the stop/poll/restart flow, attempting to call `doStart` a second time.
    *   **Flagged by:** Correctness

**Medium**

*   **Complex Orchestration Logic Leaked to the Client**
    *   The frontend is responsible for a complex multi-step process (Pre-check -> Stop -> Poll -> Start) that should be handled by the backend.
    *   This logic is fragile (e.g., if the user closes the tab) and must be duplicated by any other client. A single `/restart` endpoint on the backend would be a much better architecture.
    *   **Flagged by:** Architecture
*   **Missing UI Refresh in Pre-check Path**
    *   The new pre-check logic, which allows for an early `doStart`, fails to call `await fetchSessions()` first.
    *   The standard flow calls this function to refresh the UI. Its omission in the new path will lead to UI inconsistencies.
    *   **Flagged by:** Correctness

---

### Reviewer Disagreements

*   There were no direct disagreements. However, reviewers had different perspectives on the client-side concurrency guard (`_restartingSessions`):
    *   The **Security** reviewer praised it as a positive pattern for preventing client-side race conditions.
    *   The **Architecture** reviewer identified it as a "band-aid" made necessary by a severe backend race condition (see "Backend `/stop` Endpoint Is Not Idempotent" above).

---

### Verdict

The most critical issue to address first is the **Missing Authentication in Raw `fetch` Calls**. This is a **Blocker** that will cause the new restart functionality to fail completely in a production environment. While the architectural issues are more severe long-term, this bug prevents the feature from working at all.

---

<details>
<summary>Individual Reviews (click to expand)</summary>



## Security Reviewer

Here is a security review of the provided PR diff. 

### 1. Missing Authentication in Raw `fetch` Calls
**Severity:** **Medium**
**Category:** Auth Weakness / Functional Bug

**Description:**
The PR replaces some `apiCall` usages with raw `fetch(SINGLE_URL)` to check the session status. In many client-side applications, wrapper functions like `apiCall` are responsible for automatically injecting authentication credentials (e.g., `Authorization: Bearer <token>` headers). 
* If the API requires authentication headers, these raw `fetch` requests will fail (e.g., returning `401 Unauthorized`). Because the code ignores non-`ok` responses, the pre-check will silently fail and the polling loop will always time out, breaking the restart functionality.
* Conversely, if these raw `fetch` requests *succeed* without authentication, it indicates a server-side **Auth Bypass / Data Exposure** vulnerability, as unauthenticated users would be able to poll and read session statuses.
*(Note: If authentication is handled entirely via same-origin cookies, `fetch` will include them by default and this may function correctly, but it is still safer to use the established API wrapper).*

**Recommendation:**
Ensure that the raw `fetch` calls include the necessary authentication credentials. If `apiCall` cannot be used because it flattens HTTP errors (as noted in the comments), consider creating a wrapper around `fetch` that injects the required auth headers but returns the raw `Response` object, or manually add the headers to the `fetch` options here.

---

### 2. Potential URL Injection in `doStart`
**Severity:** **Low**
**Category:** Injection Risk

**Description:**
The developer correctly identified a potential URL injection/path traversal risk in the old code and fixed it by introducing `const encodedName = encodeURIComponent(name);` for the `/stop` and polling endpoints. This is a great security improvement. 

However, `doStart(name)` is still called using the raw, unencoded `name` variable:
```javascript
await doStart(name);
```
If the `doStart` function internally constructs a URL using this raw variable (e.g., `API + '/api/sessions/' + name + '/start'`), it will be vulnerable to the exact same path traversal/injection issue that was just fixed in `doRestart`.

**Recommendation:**
Review the implementation of `doStart(name)` to ensure it either applies `encodeURIComponent` internally, or update the calls in this function to pass `encodedName` instead (e.g., `await doStart(encodedName);`).

---

### 🛡️ Positive Security Findings (Info)
* **Concurrency Guard:** The introduction of the `_restartingSessions` Set is a solid pattern. It safely prevents race conditions and overlapping state-mutating requests (Denial of Service / state corruption) if a user spams the restart button. The `finally` block ensures the lock is released even if network errors occur.
* **URL Encoding:** As mentioned above, wrapping the user-controlled `name` variable in `encodeURIComponent(name)` before appending it to the URL path successfully mitigates client-side path traversal risks.

## Architecture Reviewer

Here is an architectural review of the provided PR diff. 

While the PR modifies frontend JavaScript (embedded within a Python string), the changes **reveal significant architectural flaws in the backend API design and state management**. The frontend is being forced to orchestrate complex workarounds for backend deficiencies.

Here are the findings, categorized and rated by severity:

### 1. API Contract Inconsistency: Semantic mismatch of the `running` field
**Rating: High**
* **Issue:** The comments explicitly state a major API contract violation: The list endpoint (`/api/sessions`) and the single-item endpoint (`/api/sessions/{name}`) return different semantic meanings for the `running` boolean. The list endpoint means "tmux session exists", while the single endpoint means "process inside tmux is running".
* **Architectural Impact:** This breaks RESTful consistency and the Principle of Least Astonishment. Clients cannot trust the API models. If a client fetches the list, a session might appear "running", but fetching the individual session shows it as "stopped".
* **Recommendation:** The backend API contract must be unified. The API should likely return two distinct fields (e.g., `tmux_active: boolean` and `process_running: boolean`) or a unified `state` enum (`"STOPPED"`, `"EXITING"`, `"RUNNING"`). Do not rely on the frontend to know which endpoint holds the "true" state.

### 2. Backwards Compatibility / State Machine: Backend `/stop` race condition
**Rating: High**
* **Issue:** The PR adds a "Pre-check" to see if the session is already stopped before calling `/stop`. The comment explains why: *"Without this, the queued `_bg_stop` worker can race a freshly-launched Claude and kill it."*
* **Architectural Impact:** This indicates the backend `/stop` endpoint is not idempotent and lacks proper state validation. If the backend receives a `/stop` command for an already-stopped session, it blindly queues a background worker that will execute later, potentially killing a *new* process started by a subsequent `/start` command.
* **Recommendation:** Fix the backend state machine. The `/stop` endpoint must be idempotent. When `/stop` is called, the backend should check the current state; if the process is already stopped, it should immediately return a `200 OK` (or `304 Not Modified`) and **not** queue a background worker. The frontend should not be responsible for protecting the backend from its own race conditions.

### 3. Missing Abstraction: Leaking orchestration to the client
**Rating: Medium**
* **Issue:** The frontend is orchestrating a complex, multi-step state machine: Pre-check -> Conditional Stop -> 30-second Polling Loop -> Start. 
* **Architectural Impact:** This is a classic "Chatty I/O" anti-pattern. If the user closes their browser tab or loses network connection during the 30-second polling window, the session will stop but never restart. Furthermore, if you ever build a CLI or another client, this complex logic must be duplicated.
* **Recommendation:** Introduce a `/api/sessions/{name}/restart` endpoint on the backend. The backend is the correct place to handle synchronous/asynchronous waiting, process lifecycle management, and ensuring that a stop is fully completed before a start is initiated. The frontend should just call `/restart` and optionally poll for the final "running" state.

### 4. Design Pattern: Embedded Frontend Code in Backend (Technical Debt)
**Rating: Low**
* **Issue:** The JavaScript is embedded as a raw string inside a Python file (`def _share_viewer_html(...) -> str:`). 
* **Architectural Impact:** While this PR didn't introduce this pattern, adding complex orchestration logic (Promises, async/await, Set-based mutexes, error handling) into a Python string makes the code unmaintainable. It bypasses JS tooling (linters, TypeScript, minifiers) and makes testing impossible without spinning up the Python server.
* **Recommendation:** Extract the HTML/JS into separate static files that the Python server reads and serves. 

### Summary
While this PR successfully patches the immediate bugs (preventing overlapping restarts and fixing the polling target), it does so by adding **frontend band-aids over backend architectural wounds**. 

**Merge recommendation:** If this is a hotfix, merge it to unblock users. However, a follow-up ticket should be created immediately to fix the backend API contract (unifying the `running` definition), make the backend `/stop` endpoint idempotent, and ideally move the `restart` orchestration to the backend.

## Correctness Reviewer

Here is a review of the PR diff, categorized by severity:

### 1. Raw `fetch` is likely missing authentication headers (Severity: Blocker)
**Issue:** The original code uses `apiCall`, which typically acts as a wrapper to inject necessary authentication headers (e.g., `Authorization: Bearer <token>`) or credentials. The new code introduces raw `fetch(SINGLE_URL)` for the pre-check and the polling loop. Given the Python function signature `_share_viewer_html(token: str, ...)`, the API almost certainly requires authentication.
**Impact:** If `fetch` is sent without the required headers, the server will return `401 Unauthorized` or `403 Forbidden`. 
- In the pre-check, `pre.ok` will be false, silently skipping the optimization.
- In the polling loop, `r.ok` will always be false. The loop will exhaust all 58 attempts (taking ~30 seconds) and ultimately fail with the *"Stop didn't take effect"* toast, completely breaking the restart functionality.
**Recommendation:** Pass the required headers into the `fetch` calls (e.g., `headers: { 'Authorization': 'Bearer ' + token }`), or update `apiCall` so it can return raw response objects/status codes and use that instead.

### 2. `doStart` errors are swallowed by the pre-check `try/catch` (Severity: High)
**Issue:** In the pre-check block, `await doStart(name);` is placed inside the `try` block intended to catch `fetch` network errors. 
**Impact:** If `doStart(name)` throws an exception (e.g., due to an API error or its own internal network failure), the `catch (e)` block will catch it and assume it was just a "network blip on pre-check". The code will then incorrectly fall through to the standard flow: it will attempt to stop the session, poll for 30 seconds, and then call `doStart(name)` a *second* time.
**Recommendation:** Move the `doStart` execution outside of the `try/catch` block.
```javascript
let shouldStart = false;
try {
  const pre = await fetch(SINGLE_URL, /* add headers here */);
  if (pre.status === 404) {
    showToast('Session no longer exists');
    return;
  }
  if (pre.ok) {
    const data = await pre.json();
    if (!data.running) shouldStart = true;
  }
} catch (e) {
  // Network blip on pre-check
}

if (shouldStart) {
  await fetchSessions(); // See finding #3 below
  await doStart(name);
  return;
}
```

### 3. Missing `fetchSessions()` in the pre-check early return (Severity: Medium)
**Issue:** At the end of the standard stop+poll flow, the code calls `await fetchSessions();` with the comment `// refresh card UI before doStart`. However, in the new pre-check early return, `await doStart(name);` is called directly without refreshing the sessions first.
**Impact:** If the UI relies on `fetchSessions()` to clear out stale state or update the visual indicators before a start is initiated, the pre-check path will cause a UI inconsistency.
**Recommendation:** Add `await fetchSessions();` right before `await doStart(name);` in the pre-check early return block to maintain consistent behavior.

### 4. Missing tests for new code paths (Severity: Low)
**Issue:** The PR introduces several new logical branches that handle specific race conditions and network states, but no tests are included in the diff.
**Impact:** Future refactors could easily break the pre-check logic or the 404 handling without CI failing.
**Recommendation:** Ensure end-to-end or integration tests are added to cover:
1. The pre-check successfully bypassing the stop/poll when a session is already stopped.
2. A `404 Not Found` response during both the pre-check and the polling loop.
3. Network errors (simulated `fetch` rejections) during polling to ensure the loop recovers and continues.
4. Concurrent calls to `doRestart` to verify the `_restartingSessions` Set correctly blocks overlaps.

</details>

kmichels:
# Review Panel: PR #27: fix(ui): doRestart never called doStart due to broken running flag

**Reviewers:** Security, Architecture, Correctness
**Date:** 2026-04-28

## Moderated Findings

Here is the consolidated and prioritized findings list based on the reviewers' feedback. 

*(Note: Findings flagged as "Low" by only a single reviewer have been dropped per your instructions. Positive findings were also excluded to focus on actionable issues.)*

### High
* **Missing timeout on raw `fetch()` calls:** Native `fetch` lacks a default timeout. If the network hangs, the `finally` block will never execute, permanently locking the session in the `_restartingSessions` state. *(Flagged by: Correctness)*
* **Inconsistent API Contracts (`running` semantics):** The list endpoint and `/info` endpoint use the exact same property name (`running`) to represent completely different states (tmux session existence vs. process state), forcing the client to handle backend implementation details. *(Flagged by: Architecture)*

### Medium
* **Missing URL Encoding in `doStart`:** While `encodeURIComponent` was correctly added for `/info` and `/stop`, the downstream `doStart(name)` function still receives the raw, unencoded `name`. This leaves a lingering risk of URL injection/path traversal or broken requests. *(Flagged by: Security, Architecture)*
* **Leaky Abstraction & Monkey-Patched `fetch`:** Bypassing the standard `apiCall` wrapper for raw `fetch` breaks error handling consistency. Furthermore, globally monkey-patching `fetch` to inject auth headers is a dangerous anti-pattern that risks leaking credentials to third-party domains.
  * *Disagreement Flag:* Security rated this as **Low** (focusing on the defense-in-depth auth leakage risk), while Architecture rated it as **Medium** (focusing on the broken HTTP abstraction and anti-pattern). *(Flagged by: Architecture, Security)*
* **Race condition with other UI actions:** The `_restartingSessions` guard prevents overlapping restarts, but it does not block a user from manually clicking "Start" or "Stop" during the 30-second polling loop, which will lead to false error toasts. *(Flagged by: Correctness)*
* **Missing test coverage:** The PR introduces significant new branching (e.g., 404s, network errors, polling exhaustion) that currently lacks end-to-end or unit test coverage. *(Flagged by: Correctness)*

### Low
* *(None — all single-reviewer Low findings were dropped)*

---

### Verdict
**Most critical issue to address first:** The **missing timeout on raw `fetch()` calls**. 
While the inconsistent API contract is a severe architectural flaw, the missing timeout introduces a direct, unrecoverable client-side Denial of Service. If a request hangs, the user is permanently locked out of restarting that session until they completely refresh the page. This should be fixed immediately by passing an `AbortSignal.timeout()` to the fetch calls.

---

<details>
<summary>Individual Reviews (click to expand)</summary>



## Security Reviewer

Based on the security review of the provided PR diff, **no new security vulnerabilities were introduced**. In fact, this PR improves the security posture of the application by fixing a pre-existing client-side injection flaw. 

Here is the detailed breakdown of findings and recommendations based on the context revealed by the diff:

### 1. URL Injection / Path Traversal (Mitigated)
* **Rating:** **Positive Finding** (Fixes a previously **Medium** risk)
* **Category:** Injection Risk
* **Description:** In the original code, the `name` parameter was directly concatenated into the URL (`API + '/api/sessions/' + name + '/stop'`). If a user or attacker could control the `name` variable, they could use path traversal characters (e.g., `../`) or URL control characters (`?`, `#`) to manipulate the API request and potentially interact with unintended endpoints. 
* **Resolution:** The PR correctly introduces `encodeURIComponent(name)` to sanitize the input before constructing `INFO_URL` and `STOP_URL`, effectively neutralizing this vulnerability.

### 2. Potential URL Injection in `doStart` (Out of Scope of Diff)
* **Rating:** **Low / Medium** (Requires verification)
* **Category:** Injection Risk
* **Description:** At the end of the `doRestart` function, `await doStart(name);` is called using the raw, unencoded `name` variable. 
* **Recommendation:** Ensure that the `doStart` function (which is outside this diff) has also been updated to use `encodeURIComponent(name)` when constructing its API requests. If it hasn't, the path traversal vulnerability mitigated in this PR still exists in the start flow.

### 3. Auth Header Leakage Risk via Monkey-Patched `fetch`
* **Rating:** **Low** (Informational / Defense-in-Depth)
* **Category:** Auth Weaknesses
* **Description:** A code comment explicitly notes: `window.fetch is monkey-patched at line 11878 to inject auth headers.` While this works for internal API calls, global monkey-patching of `fetch` can be dangerous. If the client-side application ever uses `fetch()` to request resources from an external or third-party domain (e.g., fetching an external avatar, calling a public API), the monkey-patch will blindly attach your application's authentication headers, leaking them to the third party.
* **Recommendation:** Verify the monkey-patch implementation at line 11878. Ensure it includes an origin check (e.g., `if (url.startsWith(API) || url.startsWith(window.location.origin))`) before appending the authorization headers.

### 4. State Management / Denial of Service
* **Rating:** **Low** (Informational)
* **Category:** Application Logic
* **Description:** The PR introduces a `_restartingSessions` Set to prevent overlapping polling loops. The developer correctly placed `_restartingSessions.delete(name);` inside a `finally` block. This is excellent defensive programming, as it ensures that if the network requests fail or throw an unexpected exception, the session name is still cleared from the Set, preventing a client-side Denial of Service where a user becomes permanently locked out of restarting a session.

### Summary of Requested Categories:
* **Injection Risks:** Improved (URL encoding added). Check `doStart`.
* **Auth Weaknesses:** None introduced. (See note on monkey-patched fetch).
* **Data Exposure:** None found.
* **Secrets in Code:** None found.

## Architecture Reviewer

Here is an architectural review of the provided PR diff. 

While the PR successfully fixes a race condition and improves the robustness of the restart flow, it reveals several underlying architectural issues and tech debt in both the frontend and backend.

### 🔴 High Severity

**1. Inconsistent API Contracts (Semantic mismatch of `running`)**
*   **Issue:** The comment explicitly states: *"The list endpoint reports running == 'tmux session exists'... /info uses is_running() which correctly returns false once Claude exits."* 
*   **Architectural Impact:** This is a severe API design flaw. Two endpoints returning a property with the exact same name (`running`) for the exact same resource (a session) have completely different semantic meanings. This forces the client to understand backend implementation details (tmux vs. process state) to function correctly.
*   **Recommendation:** The backend API needs to be unified. The list endpoint and the `/info` endpoint must share the same serialization logic. If both states are needed, the API should explicitly return `tmux_alive: boolean` and `process_running: boolean`.

### 🟡 Medium Severity

**2. Leaky Abstraction (`apiCall` vs raw `fetch`)**
*   **Issue:** The code bypasses the standard `apiCall` wrapper and uses raw `fetch` because `apiCall` "flattens 404 from 5xx". Furthermore, the comment notes that `window.fetch` is monkey-patched elsewhere to inject auth headers.
*   **Architectural Impact:** The HTTP client abstraction is broken. By forcing developers to drop down to raw `fetch` to read HTTP status codes, error handling becomes inconsistent. Monkey-patching native browser APIs (`window.fetch`) is also a dangerous anti-pattern that can break third-party libraries and makes the code hard to trace.
*   **Recommendation:** Refactor `apiCall` so it doesn't swallow HTTP semantics. It should either throw typed Error objects (e.g., `NotFoundError`, `ServerError`) or return a standardized wrapper object `({ data, status, error })`. Stop monkey-patching `fetch` and instead use an interceptor pattern or a dedicated HTTP client class.

**3. Missing URL Encoding in Downstream Functions**
*   **Issue:** The PR correctly introduces `encodeURIComponent(name)` for the `/info` and `/stop` endpoints. However, it passes the raw, unencoded `name` to `doStart(name)`.
*   **Architectural Impact:** If `name` contains special characters (spaces, slashes, etc.), the `/info` and `/stop` calls will succeed, but `doStart` will likely fail or target the wrong URL if it doesn't implement its own encoding.
*   **Recommendation:** Verify the contract of `doStart`. If `doStart` expects an unencoded string and encodes it internally, this is fine. If not, this is a bug. Architecturally, URL encoding should be handled by the API client layer (e.g., `apiCall`), not manually concatenated in business logic.

### 🟢 Low Severity

**4. Polling vs. Event-Driven Architecture**
*   **Issue:** The client polls the server up to 58 times over 30 seconds to check if a process has stopped.
*   **Architectural Impact:** While the 1s warmup and 500ms intervals are a pragmatic fix for now, long-polling ties up client resources and generates unnecessary backend load. 
*   **Recommendation:** For future iterations, consider moving to Server-Sent Events (SSE) or WebSockets for session state changes, or implement an asynchronous Task API where the client can await a specific job ID.

**5. Monolithic God-File Anti-Pattern**
*   **Issue:** This JavaScript is being injected via a Python string at line `14111` of `amux-server.py`.
*   **Architectural Impact:** Maintaining complex frontend logic (concurrency control, DOM manipulation, polling) inside a 14,000+ line Python backend file severely harms maintainability, prevents the use of modern JS tooling (linters, bundlers, TypeScript), and breaks separation of concerns.
*   **Recommendation:** Extract the frontend assets into separate `.js` and `.html` files and serve them statically.

**6. Potential Race Condition on "Already Stopped"**
*   **Issue:** The pre-check verifies if the session is already stopped. If it is running, it proceeds to call `STOP_URL`. If the session crashes or stops *exactly between* the pre-check and the `STOP_URL` call, `apiCall` might return an error (e.g., 404 or 500). The code handles this with `if (!stopResp) return;`.
*   **Architectural Impact:** If `stopResp` fails because the session *just* stopped, the function aborts entirely, meaning `doStart` is never called. The user clicked "Restart", but the system only stopped.
*   **Recommendation:** Ensure the backend `/stop` endpoint is idempotent. If you call `/stop` on an already-stopped session, it should return `200 OK` (or a specific code that `apiCall` doesn't treat as a fatal error), allowing the flow to continue to `doStart`.

---
### Summary of Good Patterns in this PR:
*   **Concurrency Guard:** The use of `_restartingSessions` (a `Set`) combined with a `try...finally` block is an excellent UI pattern to prevent overlapping mutations.
*   **Graceful Degradation:** Wrapping the polling `fetch` in a `try/catch` to ignore network blips ensures the restart loop doesn't break due to a momentary connection drop.

## Correctness Reviewer

Here is a review of the PR diff, categorized by severity:

### High
**1. Missing timeout on raw `fetch()` calls (Error Handling / Edge Case)**
Unlike some wrapper libraries (which `apiCall` might be), the native browser `fetch()` API does not have a default timeout. It will wait until the browser's internal network timeout is reached (which can be up to 5 minutes) if the server hangs or the network silently drops packets. 
Because the `fetch(INFO_URL)` calls are inside a `try...finally` block, if `fetch` hangs indefinitely, the function will block forever. The `finally` block will never execute, meaning `_restartingSessions.delete(name)` is never called. This will permanently lock the session, preventing the user from attempting another restart until they completely refresh the page.
* **Recommendation:** Add an `AbortSignal` timeout to the raw fetch calls.
  ```javascript
  const r = await fetch(INFO_URL, { signal: AbortSignal.timeout(5000) });
  ```

### Medium
**2. Race condition with other session actions (`doStart` / `doStop`)**
The `_restartingSessions` guard successfully prevents multiple overlapping `doRestart` calls for the *same* session. However, it does not prevent a user from manually triggering a start or stop action from the UI while `doRestart` is in its 30-second polling loop. 
If a user clicks "Start" while `doRestart` is polling, `doStart` will launch the session. The polling loop will then see `data.running === true`, continue polling until it exhausts `MAX_POLLS`, and eventually show the toast `"Stop didn't take effect, try again"`, even though the session is actually running.
* **Recommendation:** Ensure that `doStart` and `doStop` also check/respect the `_restartingSessions` guard, or ensure the UI disables these buttons while a restart is in progress.

**3. Missing tests for new code paths**
This PR introduces significant new branching and edge-case handling that currently lacks test coverage in the diff:
* Pre-check returns `404` (session gone)
* Pre-check returns `200` but `!data.running` (already stopped)
* Pre-check network error (fallback to standard stop)
* Polling returns `404` (session deleted during poll)
* Polling timeout exhaustion
* **Recommendation:** Since this is embedded JavaScript, add end-to-end tests (e.g., Playwright/Selenium) or extract the logic to a testable JS module to cover these specific state transitions and prevent future regressions.

### Low
**4. Potential unhandled promise rejections (Error Handling)**
If `doStart(name)` or `fetchSessions()` throws an error (e.g., due to a network failure or JSON parsing error), the error will bubble up and out of `doRestart`. While the `finally` block correctly ensures the lock is released, the unhandled rejection might cause console errors or silent failures depending on how `doRestart` is invoked (e.g., if it's attached to an `onClick` handler without a `.catch()`).
* **Recommendation:** Consider wrapping the `doStart` and `fetchSessions` calls in a `try...catch` block to gracefully handle and display a toast for start-up failures, or ensure the caller of `doRestart` handles promise rejections.

### Positive Notes (Correctness)
* **URL Encoding:** Wrapping `name` in `encodeURIComponent(name)` is a great fix. If session names contain spaces or special characters, the previous implementation would have resulted in malformed URLs.
* **Robust Polling:** Wrapping the polling `fetch` and `r.json()` in a `try...catch` is excellent. If the server temporarily returns a 502 or invalid JSON during the restart, it correctly treats it as a network blip and continues polling rather than crashing the loop.
* **Mutex Implementation:** Using a module-level `Set` with a `try...finally` block is the correct, thread-safe way to implement an async lock in JavaScript's event loop.

</details>

kmichels:
# Review Panel: PR #27: fix(ui): doRestart never called doStart due to broken running flag

**Reviewers:** Security, Architecture, Correctness
**Date:** 2026-04-28

## Moderated Findings

Here is the moderated summary of the reviews.

### Prioritized Findings

#### Blocker
*   **Extreme Monolith:** The frontend JavaScript is embedded directly within a Python file at line 14,111, preventing the use of standard frontend tooling and making the system unmaintainable.
    *   *Flagged by: Architecture*

#### High
*   **Brittle Authentication via Monkey-Patched `fetch`:** The global `fetch` function is monkey-patched to inject auth headers. This is a brittle anti-pattern that can leak credentials to third-party domains.
    *   *Flagged by: Architecture, Security*
*   **Inconsistent API Contract:** The `running` field means different things in the `/api/sessions` list endpoint vs. the `/api/sessions/{name}/info` endpoint, violating API consistency and likely causing bugs.
    *   *Flagged by: Architecture*
*   **Browser Compatibility Breaks Feature:** The use of `AbortSignal.timeout()` is not supported in older browsers. The feature will silently fail for those users because the error is swallowed by a broad `catch` block.
    *   *Flagged by: Correctness*

#### Medium
*   **Client Compensating for Backend Race Conditions:** The client-side code implements a lock (`_restartingSessions`) to prevent a backend background worker from killing a newly started session. This indicates a flawed concurrency model in the backend, which should be responsible for its own state management.
    *   *Flagged by: Architecture*
*   **Leaky `apiCall` Abstraction:** The `apiCall` wrapper hides important HTTP status codes (like 404), forcing the developer to bypass it and use raw `fetch`, leading to inconsistent network logic.
    *   *Flagged by: Architecture*
*   **Broad `catch` Blocks Swallow Errors:** Empty `catch (e) {}` blocks hide all errors (network, syntax, runtime), making the feature difficult to debug when it fails.
    *   *Flagged by: Correctness*
*   **Potential Missing URL Encoding:** The `doStart(name)` function is called with a raw `name`, which may re-introduce the same path injection/traversal vulnerability that was fixed elsewhere in the PR if `doStart` does not perform its own encoding.
    *   *Flagged by: Correctness, Security*
*   **Insufficient Concurrency Control:** The `_restartingSessions` lock only prevents duplicate `doRestart` calls, but a user can still click the "Start" or "Stop" buttons during a restart, leading to other race conditions.
    *   *Flagged by: Correctness*

---

### Disagreements Noted

*   **Monkey-patched `fetch`:** Rated **High** by Architecture (brittle anti-pattern) vs. **Medium** by Security (conditional token leakage risk).
*   **Missing URL Encoding in `doStart`:** Rated **Medium** by Correctness (breaks functionality for names with special characters) vs. **Low** by Security (potential injection risk).

---

### Verdict

The most critical issue to address first is the **`AbortSignal.timeout()` browser compatibility**. It is a high-severity correctness bug in the new code that will cause the feature to silently fail for a segment of users.

---

<details>
<summary>Individual Reviews (click to expand)</summary>



## Security Reviewer

Here is a security review of the provided PR diff. 

Overall, this PR improves the security and stability of the client-side code. It introduces proper URL encoding (fixing a potential path injection vulnerability) and implements safe polling with timeouts. However, there are a couple of contextual security considerations to review.

### Findings

#### 1. Auth Token Leakage via Monkey-Patched `fetch` (Auth Weakness)
**Severity:** **Medium**
**Description:** 
A comment in the diff notes: `(window.fetch is monkey-patched at line 11878 to inject auth headers.)`. 
While this is a common pattern for SPAs, globally monkey-patching `window.fetch` to blindly append authentication headers (like `Authorization: Bearer ...`) can be dangerous. If the application ever uses `fetch()` to request resources from a third-party domain (e.g., fetching an external avatar, loading a user-provided URL, or sending telemetry), the auth token will be leaked to that third-party server.
**Recommendation:** 
Ensure the monkey-patch at line 11878 explicitly checks the destination URL before appending the auth header. It should only attach credentials if the request URL is same-origin or matches the trusted `API` base URL.

#### 2. Potential Missing URL Encoding in `doStart` (Injection Risk)
**Severity:** **Low**
**Description:** 
This PR excellently introduces `encodeURIComponent(name)` to construct the `INFO_URL` and `STOP_URL`. This prevents URL/path injection if a session `name` contains special characters (like `../`, `?`, or `#`). 
However, at the end of the function, `doStart(name)` is called using the raw, unencoded `name`. If `doStart` does not internally encode the name before appending it to a URL, the same path injection vulnerability that was fixed here might still exist in the start flow.
**Recommendation:** 
Verify that the `doStart` function implementation also uses `encodeURIComponent(name)` when constructing its API endpoints.

---

### Security Improvements Noted (Positive Findings)
* **Path Traversal / URL Injection Fix:** The transition from `API + '/api/sessions/' + name + '/stop'` to using `encodeURIComponent(name)` is a great security improvement. It prevents attackers from manipulating the API path using crafted session names.
* **Denial of Service (DoS) Prevention:** The polling mechanism is implemented securely. It uses a strict upper bound (`MAX_POLLS = 58`) to prevent infinite loops, and utilizes `AbortSignal.timeout()` to ensure that hanging network requests do not permanently lock up the UI or the `_restartingSessions` state.
* **Race Condition Mitigation:** The `_restartingSessions` Set effectively prevents overlapping execution flows, which protects against unexpected state mutations and API spamming.

## Architecture Reviewer

Here is an architectural review of the provided PR diff, focusing on design patterns, API contracts, backwards compatibility, and abstractions.

### 1. Inconsistent API Contract for the `running` state
**Rating: High**
*   **Issue:** The developer comments reveal a severe API design flaw: the `/api/sessions` (list) endpoint and the `/api/sessions/{name}/info` endpoint both return a `running` boolean, but they mean completely different things. The list endpoint means "tmux session exists", while the info endpoint means "process inside tmux is running".
*   **Architectural Impact:** This violates the Principle of Least Astonishment and breaks API contract consistency. Any other client (or future developer) consuming this API will likely introduce bugs by assuming `running` means the same thing across endpoints.
*   **Recommendation:** The backend API should be updated to return distinct, unambiguous fields (e.g., `tmux_alive: boolean` and `process_alive: boolean`) across *all* session endpoints.

### 2. Client Compensating for Backend Race Conditions
**Rating: Medium**
*   **Issue:** The comment states: *"Without this [pre-check], the queued `_bg_stop` worker can race a freshly-launched Claude and kill it."*
*   **Architectural Impact:** State machine enforcement is leaking into the frontend. If the backend has a queued background worker (`_bg_stop`) that can accidentally kill a newly started session because of timing issues, the backend's concurrency model is flawed. The client-side `_restartingSessions` lock only protects against double-clicks in *one* browser tab; it won't protect against multiple users, multiple tabs, or API scripts.
*   **Recommendation:** The backend must be responsible for its own state transitions. The backend `/start` endpoint should invalidate or safely bypass any pending `_bg_stop` tasks for that specific session.

### 3. Leaky Abstraction in API Client (`apiCall` vs raw `fetch`)
**Rating: Medium**
*   **Issue:** The developer is forced to abandon the standard `apiCall` wrapper and use raw `fetch` because `apiCall` "flattens both [404 and 5xx errors]".
*   **Architectural Impact:** The `apiCall` abstraction is too rigid and hides critical HTTP semantics from the caller. This leads to fragmented network logic where developers mix and match `apiCall` and raw `fetch` depending on whether they need to read HTTP status codes.
*   **Recommendation:** Refactor `apiCall` to either return a standardized response object (e.g., `{ data, status, error }`) or throw specific Error subclasses (e.g., `NotFoundError`) so consumers don't have to bypass the abstraction.

### 4. Brittle Authentication Pattern (Monkey-patched `fetch`)
**Rating: High**
*   **Issue:** The comment casually mentions: `window.fetch is monkey-patched at line 11878 to inject auth headers.`
*   **Architectural Impact:** Monkey-patching global built-ins is a major anti-pattern. It makes the system highly brittle, complicates debugging, and can easily conflict with third-party libraries or browser extensions.
*   **Recommendation:** Implement a dedicated HTTP client abstraction (e.g., an `ApiClient` class or a configured `axios` instance) that handles auth header injection, rather than mutating the global `window.fetch` object.

### 5. Missing Abstraction: Imperative Polling Logic
**Rating: Low**
*   **Issue:** The polling logic (the `for` loop, `setTimeout`, `try/catch` for network blips, and manual timeout calculations) is highly imperative and clutters the business logic of `doRestart`.
*   **Architectural Impact:** Code duplication and reduced readability. If another component needs to poll an endpoint, this 20-line block will likely be copy-pasted.
*   **Recommendation:** Extract this into a reusable asynchronous utility function. For example: `await pollUntil(url, predicateFn, maxRetries, intervalMs)`.

### 6. Extreme Monolith / Lack of Separation of Concerns
**Rating: Blocker (for long-term maintainability)**
*   **Issue:** This diff is modifying JavaScript embedded inside a Python file (`amux-server.py`) at **line 14,111**.
*   **Architectural Impact:** Serving massive blocks of inline JavaScript from a Python string prevents the use of standard frontend tooling (TypeScript, ESLint, minifiers, bundlers, Jest). It represents a complete breakdown of the separation between backend and frontend architectures.
*   **Recommendation:** The frontend assets (HTML/JS/CSS) must be extracted into separate static files, served by the Python backend, and managed via a standard frontend build pipeline.

### Positive Architectural Notes (Things done right in this PR):
*   **Backwards Compatibility:** The addition of `encodeURIComponent(name)` is a solid fix that prevents URL routing errors for session names containing special characters, improving API robustness.
*   **Resilience:** The use of `AbortSignal.timeout()` and swallowing network blips (`catch (e) { // Network blip }`) during the polling phase makes the client much more resilient to transient network failures.
*   **Resource Cleanup:** Wrapping the client-side lock in a `try...finally` block ensures `_restartingSessions.delete(name)` always executes, preventing permanent UI deadlocks if an unhandled exception occurs.

## Correctness Reviewer

Here is a review of the PR diff, categorized by severity:

### 🔴 High
**`AbortSignal.timeout()` compatibility and silent failures**
`AbortSignal.timeout()` is relatively new (added in Safari 16, Chrome 103, Firefox 100). If a user is on an older browser, calling it will throw a `TypeError` (`AbortSignal.timeout is not a function`). 
Because your `catch (e) {}` blocks silently swallow *all* errors, the polling loop will fail instantly 58 times, wait 500ms between each, and eventually show "Stop didn't take effect" without ever actually checking the server. This completely breaks the restart feature for those users.
* **Recommendation:** Use `AbortController` with `setTimeout` for broader compatibility.
  ```javascript
  const controller = new AbortController();
  const timeoutId = setTimeout(() => controller.abort(), 3000);
  try {
    const r = await fetch(INFO_URL, { signal: controller.signal });
    // ...
  } finally {
    clearTimeout(timeoutId);
  }
  ```
*(Note: Also ensure the monkey-patched `window.fetch` mentioned in the comments correctly passes the `signal` property through to the native fetch).*

### 🟡 Medium
**Broad `catch` blocks swallow unexpected errors**
The empty `catch (e) {}` blocks in both the pre-check and polling loop swallow all errors, including typos, `ReferenceError`s, and JSON parsing errors (`SyntaxError`). This makes debugging very difficult if the server returns an HTML error page instead of JSON, or if there's a typo in the `try` block.
* **Recommendation:** At a minimum, log the error to the console so it isn't a silent failure.
  ```javascript
  } catch (e) {
    if (e.name !== 'AbortError') console.error('Polling error:', e);
  }
  ```

**Concurrency with `doStart` and `doStop` (Race Conditions)**
The `_restartingSessions` Set perfectly prevents multiple overlapping `doRestart` calls. However, if the UI buttons aren't disabled during the restart process, a user could still click "Start" or "Stop" while `doRestart` is polling. This could lead to race conditions (e.g., `doStart` being called twice concurrently).
* **Recommendation:** Consider using a more generic `_lockedSessions` Set that guards `doStart`, `doStop`, and `doRestart` for a given session name to prevent all concurrent mutations.

**`doStart(name)` might need URL encoding**
You correctly added `encodeURIComponent(name)` for the `INFO_URL` and `STOP_URL` to fix potential issues with special characters in session names. However, `doStart(name)` is still called with the raw, unencoded `name`. 
* **Recommendation:** Verify if `doStart` handles URL encoding internally. If it doesn't, it will fail for session names with special characters.

### 🟢 Low
**Polling loop worst-case duration is much longer than 30s**
The comment notes `1s warmup + 58 * 500ms ≈ 30s total`. This assumes the `fetch` request resolves instantly. If the server is hanging and `fetch` hits the 3-second timeout every time, each iteration takes 3.5 seconds. `58 iterations * 3.5s ≈ 203 seconds` (over 3 minutes) before the user sees the failure toast.
* **Recommendation:** If you want a strict 30-second timeout, track the start time and break the loop based on elapsed time rather than a fixed iteration count:
  ```javascript
  const startTime = Date.now();
  while (Date.now() - startTime < 30000) { ... }
  ```

**Missing tests for new JS logic**
The new pre-check and polling logic introduces several new branches (404 handling, already stopped, network blips). 
* **Recommendation:** Ensure there are integration tests (e.g., Playwright/Selenium) that cover these new code paths, particularly the 404 "Session no longer exists" case and the successful polling case.

</details>

kmichels:
# Review Panel: PR #27: fix(ui): doRestart never called doStart due to broken running flag

**Reviewers:** Security, Architecture, Correctness
**Date:** 2026-04-28

## Moderated Findings

Here is the consolidated and prioritized list of findings from the review panel. 

*Note: The following "Low" severity findings were dropped as they were only flagged by a single reviewer: Manual URL Construction (Architecture), Stale UI on 404 (Correctness), and Missing Tests (Correctness).*

### 🔴 Blocker
* **Backend State Machine Flaw & Lack of Idempotency** 
  * **Reviewers:** Architecture, Correctness
  * **Disagreement Flag:** Security and Correctness praised the frontend `_restartingSessions` lock as a robust fix. Architecture disagrees on its sufficiency, warning that a frontend lock only protects a single browser tab and dangerously masks the underlying backend flaw.
  * **Description:** The frontend is being forced to orchestrate backend race conditions. If a user triggers rapid `/stop` and `/start` commands (e.g., via multiple tabs or cURL), the backend's queued `_bg_stop` worker can kill the newly launched session. The backend `/stop` endpoint must be made strictly idempotent and handle state transitions safely without relying on frontend guards.

### 🟠 High
* **Monkey-Patched `fetch` & Leaky API Abstraction**
  * **Reviewers:** Security, Architecture
  * **Description:** Bypassing the standard `apiCall` to use a custom timeout wrapper fragments network logic. More critically, relying on a globally monkey-patched `fetch` to inject auth headers is a brittle anti-pattern and a security risk. If the app fetches from third-party domains, it will leak credentials. `apiCall` should be refactored to handle timeouts and errors properly without relying on globals.
* **Inconsistent API Contracts for "Running" State**
  * **Reviewers:** Architecture
  * **Description:** The `/list` endpoint returns `true` if the tmux session exists, while `/info` returns `true` only if the underlying process is running. This conflicting definition of "running" will cause downstream bugs. The API should be unified, ideally using a status enum (e.g., `STOPPED`, `STARTING`, `RUNNING`) instead of ambiguous booleans.

### 🟡 Medium
* **Polling Loop Duration Exceeds Intended Timeout**
  * **Reviewers:** Correctness
  * **Description:** The polling loop relies on a fixed iteration count (58 polls) assuming ~500ms per poll. However, network latency or timeouts (up to 3000ms per fetch) can cause each iteration to take 3.5 seconds, extending the intended 30-second timeout to over 3 minutes. This should be refactored to use a strict timestamp-based deadline.

---

### ⚖️ Verdict
The most critical issue to address first is the **Backend State Machine Flaw (Blocker)**. Relying on frontend JavaScript to protect the backend from corrupting its own state or killing healthy processes is fundamentally unsafe. The backend `/stop` and `/start` endpoints must be refactored for strict idempotency before any further frontend work is merged.

---

<details>
<summary>Individual Reviews (click to expand)</summary>



## Security Reviewer

### Security Review Summary

Overall, this PR improves the security and stability of the client-side code. It fixes a potential URL injection vulnerability and introduces better state management to prevent race conditions. However, a comment in the code highlights a potential architectural security risk regarding how authentication is handled globally.

Here is the detailed breakdown of findings based on your requested categories:

#### 1. Path Traversal / URL Injection (Fixed)
* **Rating:** **Low** (Positive Finding / Fix)
* **Description:** In the previous version of the code, the `name` variable was directly concatenated into the API URL (`API + '/api/sessions/' + name + '/stop'`). If `name` contained path traversal characters (e.g., `../`), it could have allowed a user or attacker to interact with unintended API endpoints. The PR correctly introduces `encodeURIComponent(name)`, which safely encodes the input and mitigates this injection risk.

#### 2. Potential Credential Leak via Monkey-Patched `fetch` (Auth Weakness)
* **Rating:** **Medium**
* **Description:** A comment in the PR notes: `// (window.fetch is monkey-patched at line 11878 to inject auth headers.)`. While this PR doesn't introduce the monkey-patch, it relies on it. If the monkey-patched `fetch` blindly appends authentication headers (e.g., `Authorization: Bearer <token>`) to *all* outgoing requests, it creates a risk of credential leakage. If the application ever uses `fetch()` to request resources from a third-party or untrusted domain (e.g., external images, user-provided links, or integrations), the auth token will be sent to that external server.
* **Recommendation:** Verify the implementation of the monkey-patched `fetch` (around line 11878). Ensure that it includes an origin/URL check so that authentication headers are **only** injected for same-origin requests or explicitly trusted API domains.

#### 3. Client-Side Race Conditions / Resource Exhaustion (Fixed)
* **Rating:** **Low** (Positive Finding / Fix)
* **Description:** The PR introduces the `_restartingSessions` Set to act as a module-level guard. This prevents multiple overlapping restart actions from piling up polling loops and queued commands, which previously could have led to client-side resource exhaustion or unintended server state mutations. 

#### 4. Data Exposure & Secrets in Code
* **Rating:** **None** (Safe)
* **Description:** 
  * **Secrets:** No hardcoded secrets, tokens, or passwords are introduced in this diff.
  * **Data Exposure:** Error handling is implemented safely. Network errors and timeouts are caught without exposing sensitive stack traces or raw error objects to the DOM. Toast messages use safe, static strings, avoiding DOM-based XSS risks.

### Conclusion
The changes in this specific diff are secure and represent a net-positive improvement over the previous implementation. The only actionable security recommendation is to audit the existing `fetch` monkey-patch mentioned in the comments to ensure it doesn't leak credentials to third parties.

## Architecture Reviewer

Here is an architectural review of the provided PR diff, focusing on design patterns, API contracts, abstractions, and system design. 

### 🔴 Blocker: Backend State Machine & Concurrency Leak
**Category:** Design Patterns / Missing Abstractions
**Description:** The comment explicitly states: *"Pre-check: if Claude already isn't running... skip /stop entirely. Without this, the queued `_bg_stop` worker can race a freshly-launched Claude and kill it."* 
This reveals a severe architectural flaw in the backend. The frontend is being forced to orchestrate backend race conditions. If a user calls `/stop` and then `/start`, the backend should be smart enough to either cancel the pending `_bg_stop` task, or ensure `_bg_stop` only targets the specific Process ID (PID) of the old session, not the newly started one.
**Impact:** If the frontend pre-check fails (e.g., due to a network blip), or if a user triggers this via a direct API call (cURL) instead of the UI, the backend will corrupt its own state and kill a healthy session.
**Recommendation:** Fix the backend state machine. The backend `/stop` endpoint must be idempotent and safely handle rapid `/stop` -> `/start` sequences without relying on the frontend to protect it.

### 🟠 High: Inconsistent API Contracts for "Running" State
**Category:** API Contract Changes
**Description:** The comments reveal that the `/list` endpoint and the `/info` endpoint have completely different definitions of what `running` means. `/list` returns `true` if the *tmux session* exists, while `/info` returns `true` only if the *underlying process* (Claude) is running.
**Impact:** This is a highly confusing API contract that will inevitably lead to bugs. Consumers of this API cannot trust the `running` boolean without knowing which endpoint it came from. 
**Recommendation:** Unify the API contract. If a session has multiple states, use an enum (e.g., `status: "STOPPED" | "STARTING" | "RUNNING" | "ZOMBIE"`) rather than a boolean that means different things in different contexts.

### 🟠 High: Leaky API Client Abstraction & Monkey-Patching
**Category:** Missing Abstractions
**Description:** The developer is forced to bypass the standard `apiCall` function and write a custom `_fetchTimeout` wrapper. The reasons given are:
1. `apiCall` flattens 404 and 5xx errors, making it impossible to handle "Not Found" gracefully.
2. The code relies on a global monkey-patch of `window.fetch` (at line 11878) to inject authentication headers.
**Impact:** Bypassing the API client leads to fragmented network logic. Relying on monkey-patched globals is a known anti-pattern that makes code brittle and hard to test. Furthermore, `apiCall` is failing its primary job if it swallows HTTP status codes.
**Recommendation:** Refactor `apiCall` to either return the full `Response` object or throw typed errors (e.g., `NotFoundError`, `ServerError`). Move the timeout logic (`AbortController`) inside `apiCall` so all network requests benefit from it.

### 🟡 Medium: Client-Side Mutex vs Backend Idempotency
**Category:** Design Patterns
**Description:** The addition of `_restartingSessions` (a module-level `Set`) is a good frontend UX pattern to prevent a user from double-clicking and spamming the server. However, it only protects *this specific browser tab*. 
**Impact:** If a user has two tabs open, or two different users have access to the same session, overlapping restarts can still hit the server. 
**Recommendation:** Keep the frontend guard for UX, but ensure the backend `/stop` and `/start` endpoints are strictly idempotent. If a restart is already in progress, the backend should return a `409 Conflict` or `423 Locked`, which the frontend can catch and display.

### 🟢 Low: Manual URL Construction
**Category:** Missing Abstractions
**Description:** The code manually concatenates strings to build URLs: `API + '/api/sessions/' + encodedName + '/info'`. 
**Impact:** While `encodeURIComponent` was correctly added in this PR to fix potential injection/routing bugs with special characters in session names, manual string concatenation is prone to missing slashes or double slashes.
**Recommendation:** Use the native `URL` object to construct paths, or add a URL builder utility to your API client abstraction.

---

### General Architectural Note (Contextual)
The diff header indicates this is happening in `amux-server.py` at line **14,111**. Serving massive blocks of inline JavaScript from a 14,000-line Python monolith is a significant maintainability risk. While outside the strict scope of this specific diff, the team should strongly consider decoupling the frontend assets (HTML/JS/CSS) from the Python backend server to allow for proper JS tooling, linting, and modularization.

## Correctness Reviewer

Here is a review of the PR diff, categorized by severity.

### Medium
**Polling loop duration can significantly exceed the intended 30 seconds**
The `MAX_POLLS` logic assumes each iteration takes ~500ms. However, if the network is slow or hanging, `_fetchTimeout` can take up to 3000ms before aborting. This means each iteration could take 3.5 seconds, causing the 58 iterations to take over 3 minutes (203 seconds) instead of the intended 30 seconds. 

*Suggestion:* Use a deadline-based approach instead of a fixed number of iterations to ensure the timeout is strictly respected regardless of network latency.
```javascript
const deadline = Date.now() + 30000; // 30 seconds from now
while (Date.now() < deadline) {
  try {
    // ... fetch logic ...
  } catch (e) {
    // Network blip
  }
  await new Promise(r => setTimeout(r, 500));
}
```

### Low
**Stale UI when a session no longer exists (404)**
If the pre-check or the polling loop encounters a 404 (`sessionGone`), the function shows a toast and returns early. However, it does not call `fetchSessions()`. This means the UI may continue to display the deleted session until the next global auto-refresh occurs.

*Suggestion:* Call `await fetchSessions();` before returning in the 404 cases so the UI immediately reflects that the session is gone.

**Missing tests for new branching logic**
This PR introduces significant new branching logic in `doRestart` (pre-check, 404 handling, already-stopped handling, timeout handling). There should be frontend unit tests or E2E tests added to cover these new code paths (especially the network error and timeout scenarios) to prevent future regressions.

**Edge Case: Session stops naturally between pre-check and `/stop`**
If the session stops naturally in the few milliseconds between the pre-check and the `apiCall(STOP_URL)`, the backend `/stop` endpoint might return an error (e.g., 404). If `apiCall` treats this as a failure and returns falsy, the restart process will abort, even though the user ultimately just wanted the session to be running. 
*Note:* If your backend `/stop` endpoint is idempotent and returns a 200 success even if the session is already stopped, you can ignore this.

### Positives / Correctness Notes
* **Race Conditions:** The use of the `_restartingSessions` Set with a `try/finally` block is an excellent and robust way to prevent overlapping restart clicks. Even if an unexpected error is thrown, the lock is guaranteed to be released.
* **URL Encoding:** Wrapping `name` in `encodeURIComponent(name)` is a great addition that prevents routing errors if session names contain special characters or spaces.
* **Timeout Implementation:** `_fetchTimeout` is implemented perfectly. Using `AbortController` ensures broad browser compatibility, and chaining `.finally(() => clearTimeout(t))` prevents memory leaks from lingering timers.

</details>

esteininger:
Applied the core fixes from this PR directly to main in fa0af19:

- **Python fix**: Excluded U+276F (❯) from `_claude_ui_visible` dingbat range — prevents false-positive `is_running()` after /exit
- **JS fix**: Rewrote `doRestart` to poll `/info` endpoint (correct `is_running()` semantics) instead of the list endpoint (tmux-exists), with deadline-based 30s timeout, `AbortController` for Safari compat, pre-check for already-stopped sessions, and `_restartingSessions` Set guard

Verified end-to-end via Playwright: full restart cycle completes in ~22s with correct stop detection and relaunch.

Review panel feedback addressed:
- AbortSignal.timeout() → AbortController + setTimeout (Safari 16 compat)
- Fixed iteration count → wall-clock deadline
- fetchSessions() added to pre-check path
- encodeURIComponent for URL safety

Thanks for the thorough investigation and fix @kmichels!

PR #41, opened 2026-06-15
fix(send): send text literally before Enter with a short delay

tsukimiya:
## Problem

In `amux send <name> <text>`, the current implementation sends the text and `Enter` in a single `tmux send-keys` call:

```bash
tmux send-keys -t "=$tname" "$text" Enter
```

Claude Code's TUI buffers/debounces input. When both arrive in one call, `Enter` can be processed before the text, so the prompt shows the typed text but is never submitted.

## Fix

Send the text literally first, wait 50ms, then send `Enter` as a separate key:

```bash
tmux send-keys -t "=$tname" -l "$text"
sleep 0.05
tmux send-keys -t "=$tname" Enter
```

This mirrors the server-side `send_text()` behavior in `amux-server.py`.

## Dependency

This branch is based on #40 (`fix-tmux-exact-session-match`) because it modifies the already-fixed `tmux send-keys -t "=$tname"` line. Please review/merge #40 first, then retarget this PR to `main`.

## Verification

- `bash -n amux` passes.
- Only `amux` is modified (8 insertions, 1 deletion).

### DIFF
--- a/amux
+++ b/amux
@@ -458,7 +458,14 @@ cmd_send() {
   else
     local text="$*"
     [[ -n "$text" ]] || die "usage: amux send <name> <text>  or  amux send <name> --keys 'C-c'"
-    tmux send-keys -t "=$tname" "$text" Enter
+    # Send text literally (-l), then Enter as a separate key after a short
+    # delay. Claude Code's TUI buffers/debounces input; sending text and Enter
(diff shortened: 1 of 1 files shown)


esteininger:
Merged manually after #40 landed — applied the send-keys split fix directly to avoid the conflict. Full credit to @tsukimiya.

PR #65, opened 2026-07-30
Add MiniMax provider to the session-aware Look up inference path

octo-patch:
Reason: A MiniMax-backed session was treated as an unknown provider in the Look up path and fell back to Claude instead of using MiniMax.

## What changed

The session-aware **Look up** inference path (`/api/lookup`) rides whatever
provider the peeked session already runs on, using that provider's lightest
model for a cheap, snappy answer. It had explicit handling for the existing
providers but none for MiniMax, so a MiniMax-backed session fell through to the
Claude fallback.

This adds first-class MiniMax handling in `amux-server.py`:

- Register `MiniMax-M2.7` as the Look up model for the `minimax` provider in
  `_PROVIDER_LOOKUP_MODEL` (the lighter of the two models — a cheap, snappy
  shot, consistent with the per-provider intent).
- Add `_minimax_region()` / `_MINIMAX_OPENAI_BASE` and
  `_minimax_session_endpoint()`, which resolve the OpenAI-compatible endpoint
  from the session's configured base URL: the China endpoint
  (`api.minimaxi.com`) when the session points there, otherwise the global
  endpoint (`api.minimax.io`). Auth reuses the session's own token
  (`ANTHROPIC_AUTH_TOKEN`), falling back to `MINIMAX_API_KEY`.
- Handle `provider == "minimax"` in `_lookup_via_provider()` by calling the
  OpenAI-compatible `chat/completions` endpoint over HTTP (MiniMax ships no
  first-party CLI). Any failure falls through to the existing Claude fallback,
  exactly like the other provider branches.

No session-creation or UI surfaces are changed; this mirrors how the existing
non-whitelisted provider is already handled in the Look up path only.

## Checks

- `python3 -c "import ast; ast.parse(open('amux-server.py').read())"` — passes
  (the CI syntax gate).
- `python3 -m pytest tests/ -q` — 156 passed, 19 skipped, including the new
  `tests/test_minimax_lookup.py` (4 tests covering model registration and
  region resolution for the global and China endpoints).

### DIFF
--- a/amux-server.py
+++ b/amux-server.py
@@ -1758,6 +1758,7 @@ _PROVIDER_LOOKUP_MODEL = {
     "claude": "haiku",
     "gemini": "gemini-2.5-flash-lite",
     "codex":  "gpt-5-mini",
+    "minimax": "MiniMax-M2.7",
 }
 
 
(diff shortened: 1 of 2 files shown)


esteininger:
Thanks for the contribution — the implementation is technically clean and the tests are well-structured. We're declining this one because we don't have users running MiniMax sessions with amux today, and we prefer to add provider support in response to confirmed user demand rather than proactively. If MiniMax becomes a requested provider, we'd revisit this pattern. Feel free to open an issue first if you'd like to discuss.

PR #76, opened 2026-08-02
fix(ci): the JS syntax gate checked one script block out of eight

Dygreens:
## What & why

`checks.yml` extracts the dashboard JS with:

```python
m = re.search(r'<script>\s*\n(.*?)</script>', content, re.DOTALL)
```

**`re.search` returns the first match.** That is **2,380 characters**. `amux-server.py` holds roughly **1.27M characters** of inline client code across **eight** blocks — including the 1,048,332-char main dashboard script and a 210,709-char second block. Everything past the first has been unparsed on every PR this repo has taken.

### Demonstrated, not argued

Break `_graphUpdateStats` in the main dashboard script and run the current CI step verbatim:

```
old check: JS OK — PASSES with the dashboard broken
```

Injecting a syntax error into each block in turn:

| | caught |
|---|---|
| Old step | **1 of 8** |
| This PR | **8 of 8** |

### Two more things now fail closed

1. The old step answered `WARNING: no <script> block found` with **exit 0**. A broken extractor read as a pass — the worst failure mode, because it needs no bad code to trigger. Now exits 1.
2. The check is a committed script (`scripts/check-inline-js.py`) rather than a heredoc, so the identical logic in `.claude/check-and-commit.sh` can call it once #66 lands. Two copies of one regex is how these drifted apart to begin with.

### Why a naive "scan all blocks" doesn't work

Probably why the original stayed narrow. Two things look like a script block and aren't:

```python
re.sub(r"<script\b.*?</script>", "", frag)     # a Python regex literal
```
Excluded by requiring a well-formed opening tag, so `<script\b` doesn't qualify.

```python
f'<script>window._AMUX_AUTH_TOKEN={_json.dumps(AUTH_TOKEN)};'   # a template
```
Only valid JS *after* Python formats it. Checked with each interpolation stubbed as `0`, which still catches a dropped brace or paren without pretending to know the values.

So the walk is over the **AST**, not the text: `ast.Constant` strings are literal, `ast.JoinedStr` are templates.

### The remaining gap is printed, not hidden

Three blocks whose `<script>` and `</script>` live in *different* string literals can't be reached verbatim by either path. The script names them with file and line on every run:

```
JS OK — 8 inline script blocks, 1,265,663 chars checked
  note: 17 chars at amux-server.py:18666 span string boundaries — not verified verbatim
  note: 1093 chars at amux-server.py:53348 span string boundaries — not verified verbatim
  note: 324 chars at amux-server.py:53424 span string boundaries — not verified verbatim
```

That last one is the service-worker registration — real JS, currently unverified. Naming it is the point: a gap nobody mentions is exactly how the original survived this long. Ethos rule 7, applied to the fix as well as the bug.

## How I tested

```
clean amux-server.py  -> exit 0   (8 blocks, 1,265,663 chars)
broken dashboard      -> exit 1
old step, same file   -> "JS OK"
```

Exit codes captured without a pipe (`$?` after `| tail` reports tail's status — I got that wrong on the first pass and re-ran it).

- Injected a syntax error into all 8 blocks individually — caught every time
- `checks.yml` parses as YAML; step list unchanged apart from the replaced one
- Verified the two false-positive classes above are excluded on clean `main`

## Checklist

- [x] **One focused change** — CI gate only; `amux-server.py` untouched
- [x] **It still parses:** `python3 -c "import ast; ast.parse(open('amux-server.py').read())"`
- [x] **Client change?** none — no `APP_VER` / `CACHE` bump needed
- [x] **Tested end-to-end** — ran the real step against clean and deliberately broken files
- [x] **Single-file rule respected** — `amux-server.py` not modified; new file is tooling under `scripts/`, alongside `watchdog.py`
- [x] Rebased on latest `main`

## Follow-up, not included here

`.claude/check-and-commit.sh` carries the **same** `re.search` bug. It's not fixed in this PR because #66 already rewrites that file and the two would conflict. Once both land, that hook should become a call to `scripts/check-inline-js.py` — happy to push that as a third PR or fold it into #66, whichever you prefer.

https://claude.ai/code/session_01VtKaMqcRLDNJ7koDCqcyTC

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -19,22 +19,14 @@ jobs:
         with: { node-version: '22' }
       - name: Python syntax (amux-server.py parses)
         run: python3 -c "import ast; ast.parse(open('amux-server.py').read())"
-      - name: JS syntax (inline dashboard script parses)
-        run: |
-          python3 - <<'PY'
-          import re, subprocess, sys, tempfile, os
-          content = open('amux-server.py').read()
-          m = re.search(r'<script>\s*\n(.*?)</script>', content, re.DOTALL)
-          if not m:
-              print('WARNING: no <script> block found'); sys.exit(0)
-          fd, path = tempfile.mkstemp(suffix='.js')
-          os.write(fd, m.group(1).encode()); os.close(fd)
-          r = subprocess.run(['node', '--check', path], capture_output=True, text=True)
-          os.unlink(path)
-          if r.returncode != 0:
-              print(r.stderr); sys.exit(1)
-          print('JS OK')
-          PY
+      # EVERY inline script block, not just the first. The inlined version of
+      # this step used re.search, which returns one match, so it validated
+      # 2,380 of ~1.27M characters of client code and stayed green with the
+      # main dashboard script broken. It lives in scripts/ rather than inline
+      # so the local PostToolUse hook can call the SAME code once #66 lands —
+      # a second copy of this logic is how the two drifted in the first place.
+      - name: JS syntax (every inline dashboard script parses)
+        run: python3 scripts/check-inline-js.py amux-server.py
       - name: Tests
         run: |
           pip install -q pytest psutil
--- a/scripts/check-inline-js.py
+++ b/scripts/check-inline-js.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+"""Syntax-check the dashboard's inline <script> blocks.
+
+The check this replaces was:
+
+    m = re.search(r'<script>\\s*\\n(.*?)</script>', content, re.DOTALL)
+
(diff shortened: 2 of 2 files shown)


esteininger:
Thanks for this — closing as obsolete rather than on its merits. The premise (~1.27M chars of inline client JS across 8 `<script>` blocks in `amux-server.py`) no longer exists: the single-file Python server was removed in the Rust migration (commit `792ce1f`, 2026-08-09), and the dashboard JS is now real static files under `crates/amux-dashboard/static/` (`app.js`, `sw.js`), syntax-gated by `node --check` in both the PostToolUse hook and CI. The `re.search`-first-block bug this fixes can't recur against static files. If you spot a gap in the current JS gate, a fresh PR against `crates/` is very welcome.

PR #77, opened 2026-08-02
fix(resume): repair amux session resume — it has never once fired

Dygreens:
## The bug

Every amux session restarts with no memory of prior work. The recovery path — the peek UI's "Load log" button — tells the session to read `~/.amux/logs/.plain/<name>.log`, which for one session is **2.6 MB / 59,900 lines**. That does not fit in a context window, so recovery does not work.

But the log is a workaround for a door that was never supposed to be locked. `start_session` already has a resume path (`claude --resume <uuid>`). **It has never executed.** From `~/.amux/serve.log`, every session on this install:

```
[start] Amux-gtm: fresh start (new session)
[start] Amux: fresh start (new session)
[start] amux-helper: fresh start (new session)
... all 8
```

`Amux-gtm.meta.json` shows `start_count: 13` — thirteen restarts, thirteen blank slates.

## Three compounding bugs, each fatal alone

| # | Bug |
|---|---|
| 1 | `cc_session_name` was persisted **only in `stop_session()`**. Any ending that wasn't a graceful stop — crash, reboot, sleep, server restart — lost it. |
| 2 | The name lookup read **only line 1** of each conversation JSONL. Claude Code writes `custom-title` on **line 2**. Measured: 0 matches across all 26 files in the project. |
| 3 | The lookup required **exactly one** match. Six files share the title `Amux-gtm` — one per fresh start. |

Bug 3 is the death spiral: every blank start created another identically-named conversation, making the next lookup *more* ambiguous. It could never self-correct.

A fourth, latent: the old lookup returned a `sessionId` read from line 1 rather than the filename stem — wrong even when it matched.

## The fix

- **Derive the name instead of storing it.** amux always launches with `--name <session name>`, so it was knowable all along. Removes the crash window entirely and repairs existing sessions with no migration.
- **Scan the JSONL header block** (bounded, 30 lines) instead of line 1.
- **Newest match wins**, with snapshot-only files filtered out (`--resume` exits on those) and a deterministic `(-mtime, stem)` tie-break.
- The decision now lives in a pure `_resume_strategy(meta, name, work_dir)` helper, because `start_session` is ~700 lines with tmux side effects and none of this was testable before.

## Two regressions this fix would have caused, caught in review and fixed here

- **`reset_session` and "New conversation" became silent no-ops.** Both worked by *deleting* meta keys — which only forced a fresh start while the lookup was dead. With the name derived, deletion accomplished nothing: the next start resumed the conversation the user had just abandoned, while reporting `"reset — fresh conversation, lane intact"`. Since `reset` exists for the out-of-context emergency, this removed the only remedy for the exact case it's for. Fixed with a positive `cc_fresh_after` watermark.
- **Title-matching pre-empted the hook-reported `cc_conversation_id`**, which `amux-server.py` documents as "the ONLY deterministic link from an amux session to its conversation JSONL" — and the not-found path *deleted* it. Now consulted before any fallback to a blank start.

Also fixed: `_session_jsonl_path_uncached` had the identical line-1 bug, so peek transcript resolution silently fell through every time.

## Evidence

Replaying the fixed decision against all 8 real sessions:

```
AI---Work-Course                     RESUME  uuid=54970364-…
Amux-gtm                             RESUME  uuid=642f49e4-…
Amux-inspector                       RESUME  uuid=5a2ede61-…
Amux                                 RESUME  uuid=126a7189-…
Family-Command-and-Control-Center    RESUME  uuid=4f89a0e2-…
Kids-Daily-Prep                      RESUME  uuid=529d8af9-…
amux-helper                          RESUME  uuid=d5baa440-…
windsor-suits-rebuild                fresh   (malformed CC_DIR — a file:// URL, pre-existing)
duplicates: none
```

**7/8 resume; 0/8 did before. No two sessions resolve to the same conversation.**

Tests: **236 passing** (178 before this branch), 30 new in `tests/test_session_resume.py` pinning each named regression — line-2 titles, the ambiguity spiral, newer-snapshot-must-not-beat-older-real-conversation, post-reset freshness, and cross-session ownership.

## Known limitation

The `int(time.time())` reset watermark assumes sub-second filesystem mtime granularity (true on APFS and ext4). On a 1-second-granularity filesystem it can cost one spurious extra fresh start immediately after a reset, self-healing on the next restart. Truncating down with `<=` is deliberate: it errs toward an extra fresh start rather than resuming an abandoned conversation.

## Scope

Part A of `docs/superpowers/specs/2026-07-31-session-context-recovery-design.md`. Part B (a compact resume brief for `/clear` and compaction) and Part C (retiering the Load log prompt) are deferred.

**Not yet done: live acceptance.** This needs deploying to a running server and restarting a session to confirm `serve.log` reports `resume=` instead of `fresh start`.

https://claude.ai/code/session_013iNkfTyW4TdD5pbwLMGibZ

### DIFF
--- a/amux-server.py
+++ b/amux-server.py
@@ -3277,6 +3277,19 @@ def _cc_session_id_for_name(session_name: str, work_dir: str) -> str:
     return candidates[0].stem if candidates else ""
 
 
+def _resolve_cc_session_name(meta: dict, name: str) -> str:
+    """Return the Claude-side session name for an amux session.
+
+    amux always launches Claude with `--name <amux session name>`, so this is
+    derivable and never needed persisting. It used to be read from meta alone,
+    and meta was only written in stop_session() — so any ending that was not a
+    graceful stop lost it and forced a fresh start on a resumable session.
+
+    A name persisted in meta still wins, so `/rename` inside Claude is honoured.
+    """
+    return meta.get("cc_session_name") or name
+
+
 # Per-session token cache — refreshed every 30s, keyed by resolved dir
 _token_cache = {"data": {}, "timestamps": {}, "time": 0}
 _TOKEN_CACHE_TTL = 120
@@ -15449,7 +15462,7 @@ def start_session(name: str, extra_flags: str = "", _skip_conv_id: bool = False)
         provider = cfg.get("CC_PROVIDER", "claude").strip().lower()
         _uuid_re = re.compile(r'^[0-9a-fA-F-]{36}$')
         if not _skip_conv_id and provider == "claude":
-            cc_session_name = meta.get("cc_session_name", "")
+            cc_session_name = _resolve_cc_session_name(meta, name)
             conv_id = meta.get("cc_conversation_id", "")
             if cc_session_name and _validate_cc_session_name(cc_session_name):
                 _sid = _cc_session_id_for_name(cc_session_name, work_dir)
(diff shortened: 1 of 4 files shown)


esteininger:
Thank you for the depth here — the diagnosis is valuable. Closing because it targets `amux-server.py`, which was removed in the Rust migration (commit `792ce1f`, 2026-08-09), so the +1725 diff can't apply to the current tree. Session start/resume now lives in the Rust server (`crates/amux-server`). If resume still isn't firing there, that underlying bug is well worth fixing — please refile against `crates/` and I'll help route it. The design/plan docs you added would be useful re-added under `docs/`.

PR #83, opened 2026-08-06
feat(board): a worker that stops on an unanswered question parks its own card

Dygreens:
## The loss this prevents

A design was approved, the session asked *"does this look right before I write it up as a spec?"*, the conversation moved to another topic, and the workstream became invisible.

**Nothing was lost for want of detection.** `AMUX-6`'s own history:

```
22:38  capture: session prompt
22:39  Advance-nudge SKIPPED — captured chat prompt, not a unit of work;
       split it into real cards. A session cannot honestly finish this.
22:39  Advance-nudge SKIPPED — (same)
22:43  Advance-nudge SKIPPED — (same)
22:47  commit 332f137 — chore(hooks): announce staleness at session start
22:49  Advance-nudge SKIPPED — (same)
```

amux noticed **five times** and said exactly the right thing — into a card log neither the human nor the agent ever opens. *A tag in a store the reader never opens is the same failure as no tag* (ethos rule 4). Note `22:47`: an unrelated commit folded onto the card too.

## The primitive already existed

The board contract has always said a card can be `doing` **and** blocked-on-human, and that `amux board needsyou <ID> "the question"` records the ask into Focus / `is:needsyou`. It depends on the agent remembering to run it — and remembering is what failed. So this adds no new surface; it makes the **existing** marker automatic.

## How

Both edges are already reported by hooks that exist, so no new scraper and no new poll:

- `Stop` → `idle`: if the last assistant turn ended on a question and the worker holds a `doing` card → tag `needs:you`, write the question into the history
- `UserPromptSubmit` → `active`: clear it

Only an **auto-applied** tag is auto-cleared. A park a session set deliberately is that session's judgment and is left alone (rule 8). `added_at` is set on the tag because the >3d re-nag times the ask from the tag, not the row (AC-178).

Computed, not model-judged (rule 2) — "did this turn end on a question mark" is string manipulation, and `idle` is the most frequent event in the log (161 `session.working` rows per 500 sampled).

## The test found a real bug first

The `?` check ran **before** markdown was stripped, so a **bold question** — which ends in `**`, and is the most common shape a question takes in a rendered answer — was missed entirely.

Seven cases, loaded from the shipped source via AST rather than replicated: ends-on-a-question, ends-on-a-statement, human-already-replied, system-reminder-is-not-an-answer, question-inside-a-code-fence, bold-question, tool-only-final-turn.

## Known limitation, surfaced rather than hidden

This session's `AMUX_SESSION` is `Amux` while its live pane is `Amux---Main` — a resumed transcript carries the identity of the pane it started in. `POST /api/sessions/Amux/report` returns **`session 'Amux' not found`**, so the Stop-hook self-report is currently inert for this session. Under that drift every check here finds nothing forever, which is indistinguishable from "no open loops" — so it now logs that it is inert instead of staying silent. Fixing the drift itself is a separate change and is called out for a decision rather than guessed at.

https://claude.ai/code/session_01MDuBnmfziM6uGvdxn2y3Lo

### DIFF
--- a/amux-server.py
+++ b/amux-server.py
@@ -4969,6 +4969,176 @@ def _render_session_transcript(name: str, max_chars: int = 40000) -> str:
     return text
 
 
+# Marker written to a card's history when the needs:you tag was applied BY the
+# idle detector rather than by a session deciding to park the card. Only an
+# auto-applied tag is auto-cleared when the human answers — a tag a session set
+# deliberately is that session's to clear, and silently dropping it would be an
+# agent overruling a judgment that was not its own (ethos rule 8).
+_AUTO_NEEDSYOU = "AUTO-NEEDSYOU"
+
+
+def _closing_question(name: str) -> str:
+    """The question the last assistant turn ended on, or "" if that turn did
+    not end by asking the human something.
+
+    Deliberately COMPUTED, not model-judged (ethos rule 2). "Did this turn end
+    on a question mark" is string manipulation; paying a model call on every
+    idle transition — the most frequent event in the log, 161 session.working
+    rows in a 500-row sample — is exactly the per-call waste that earned the
+    board-title call its throttle.
+
+    Returns "" unless the LAST conversational entry is from the assistant: if
+    the human has already spoken since, there is no unanswered question.
+    """
+    path = _session_jsonl_path(name)
+    if not path:
+        return ""
+    last_role, last_text = "", ""
+    try:
+        for o in _iter_jsonl_tail(path, max_bytes=2_000_000):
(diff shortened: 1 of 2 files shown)


esteininger:
Thanks — this is a genuinely good idea (a workstream that stops on an unanswered question parking its own card). Closing only because it targets the now-deleted `amux-server.py` (Rust migration, commit `792ce1f`, 2026-08-09); the board now lives in `crates/amux-server`. If this "open-loop needsyou" behaviour isn't already covered by the board status-request flow, it's worth reimplementing against `crates/` — I'd gladly take that PR.

PR #89, opened 2026-08-13
feat(peek): Workflow tab — a worker's setup derived from state, not from the prompt (AMUX-39)

Dygreens:
Adds a **Workflow** tab to peek: a mermaid diagram of what a worker is actually set up to do, plus flags for the parts that will not work.

`GET /api/sessions/{name}/workflow` → `{mermaid, flags, derived_from}`.

Supersedes #88 (same commit, rebased onto current main to clear a version-bump conflict).

## The one design decision

The request was a diagram to "spot check the workflow" after configuring a worker. The obvious build renders **the prose you typed**. That fails ethos rule 7: a picture drawn from the same sentence that configured the worker cannot disagree with it. It is a check that cannot fail — and worse than useless, because it renders a misconfiguration as a tidy diagram and confirms it, on the surface you opened specifically to catch that.

So every node is derived from **resolved state**: `schedules` rows (`schedules.session` already binds them per worker), the worker env, board counts, `CC_TAGS` scope. The diagram may disagree with what you believed you set up, and that disagreement is the product. Flags render above the diagram for the same reason — a diagram that only draws correct config is decoration.

## It fires on real data today

Lane `amux` returns *"No MCP servers reach this worker (CC_MCP is unset)"*. That is ethos rule 1's flagship failure — six servers in `mcp.json`, 0 of 101 sessions ever received them, invisible for months — and it is still true on this machine: 6 session envs, 0 with `CC_MCP`.

## What browser verification caught

mermaid 11 strips `<br/>` from labels under `securityLevel:'strict'` (measured BR-LOST both with and without `flowchart.htmlLabels`). Two-part labels silently concatenated: `orphan job` + `once` rendered as **"orphan jobonce"**, and the worker name ran into its path.

Nothing failed. No error, no console warning — it just quietly read wrong, which no unit test would have caught. Fixed with a separator rather than by relaxing `securityLevel`: these labels carry schedule titles straight out of the DB, and strict sanitisation is what stands between that and the SVG.

## Tests

The emitter is split into a pure `build_graph` so tests drive the real one, not a paraphrase. Both directions, because a flagger that always fires is as broken as one that never does:

- correctly-configured worker → zero flags, no `classDef warn`
- disabled schedule + enabled-but-unstartable → both flag, exactly those two nodes styled
- missing MCP flags; present MCP does not
- no `<br/>` emitted anywhere (guards the whole emitter, not two call sites)

14 pass. Board counts cross-checked against real cards. Re-rendered in-browser after the fix: 2/2 diagrams OK, 0 console errors.

## Notes for the reviewer

- **Not verified:** the click-through inside the live SPA. The dev server wedges ~8s after boot on `main` — AMUX-35 (unbounded `du` parking every tokio worker), fixed on `fix/autofix-du-wedges-server` and not yet merged — too short for a page load. Verified at both ends instead: endpoint payload against real data, and those exact payloads rendering in a real browser. The wiring between them is verified structurally (registered in `PEEK_TABS`, panel ids present, every called identifier resolves) but not clicked.
- mermaid loads from jsdelivr, the same CDN as the 22 libraries `index.html` already pulls — established pattern, not a new dependency decision.
- Not added to the `peekHiddenTabs` default, so it ships visible (ethos rule 1).
- No model call: nodes and edges are computed from rows (ethos rule 2).
- `APP_VER` and `sw.js` `CACHE` bumped together, per CLAUDE.md.
- While testing I hit four bare `serde_json::Map` indexes that panic the rate-limit sweep on any lane missing the key. I wrote a fix, then found `fix/autofix-du-wedges-server` **already has the identical `meta_i64` fix**, better documented. Dropped mine rather than ship a duplicate that would conflict with that PR. Flagging it so it is not lost if that branch stalls — `main` still has the bug.
- Separately filed AMUX-42: the pre-push guard invents foreign commits on an existing branch after merging main (`remote_sha..local_sha` is unbounded by origin, while the new-branch path directly above it correctly uses `--not --remotes=origin`). That is why this is a fresh branch rather than a rebase of #88.

### DIFF
--- a/crates/amux-dashboard/static/app.js
+++ b/crates/amux-dashboard/static/app.js
@@ -3827,6 +3827,7 @@ const PEEK_TABS = [
   { id: 'steering',   label: 'Steering' },
   { id: 'schedules',  label: 'Schedules' },
   { id: 'scope',      label: 'Scope' },
+  { id: 'workflow',   label: 'Workflow' },
   { id: 'messages',   label: 'Messages' },
   { id: 'dictation',  label: 'Dictation' },
   { id: 'issues',     label: 'Board' },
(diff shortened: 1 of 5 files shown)


esteininger:
Closing for a terminal state. CI 'check' AND 'e2e' are FAILING and the branch CONFLICTS with main (app.js/index.html/sw.js/mod.rs all actively rewritten; only workflow.rs is new). The Workflow-tab idea is sound and the ethos rule-7 framing (derive the diagram from STATE, not from the prompt that configured the worker) is right — but a red + conflicting PR can't merge and it branches from a stale main. Re-land on current main if wanted; the new workflow.rs is the salvageable core. (Already superseded #88.)

PR #91, opened 2026-08-13
fix(commit-report): attach to the card the commit NAMES, not the freshest row (AMUX-40)

Dygreens:
`commit-report` picked its attach target with **no relevance test of any kind**:

```sql
SELECT id FROM issues WHERE session=? ... AND status IN ('doing','review')
AND owner_type='agent' ORDER BY updated DESC LIMIT 1
```

Whatever the lane touched most recently won. AM-5 reported the consequence: a CLAUDE.md docs fix landed on AMUX-11, a decision card awaiting the owner, which was merely the freshest `review` row. Diagnosis and dead hypotheses live on AM-5; this is the fix.

## My own spec for this fix was wrong

The card said *"290 of the last 300 commits (96.7%) name a card id"* and proposed matching on the commit message. That number is for the message **anywhere**. This endpoint receives only `sha` and `subject`. Re-measured over the same 300 commits:

| where | hits |
|---|---|
| subject only | **80 / 300 (27%)** |
| anywhere | 293 / 300 (98%) |

Matching on `subject` would have covered a quarter of commits while looking like it covered everything — a silent under-match that reads as success. So the body is read here from the sha the caller already sends.

## Resolution order

1. A card the commit **names**, owned by this lane and still open, in the commit's own mention order.
2. No mention → a **single** `doing` card, only if unambiguous. `review` is dropped deliberately: a card in review awaits someone else, so an attach there is phantom progress on a card this lane cannot advance — the exact shape AM-5 reported.
3. Otherwise **attach nothing**. A wrong link in the append-only log is worse than an absent one, because the board contract designates that log the durable half and everything downstream trusts it.

Also: the response's `attached` is now the id the **write** chose, handed back rather than re-derived. The old code ran a *different* SELECT to build the response, so a response could name a card the write never touched — a view disagreeing with its own mechanism.

## Tests

Built from the specimen, not a convenient case — a commit whose id is **only in the body**, which is precisely what a subject matcher misses. Plus controls a parser that always returns something would fail:

- commit naming nothing → nothing
- unreadable dir / bad sha → empty, not a panic
- lowercase `amux-9` and one-letter `x-1` do not match

910 lib tests and 25 `board_api` tests pass.

## Note

No `APP_VER` bump — server-side only, no client asset changed.

### DIFF
--- a/crates/amux-dashboard/static/app.js
+++ b/crates/amux-dashboard/static/app.js
@@ -1865,6 +1865,13 @@ async function _apiErrText(r) {
         // Gate refusals carry the exact command that WOULD work — that is the
         // most actionable thing in the whole response; surface it.
         if (msg && j.cli) msg += ' — try: ' + j.cli;
+        // Same idea, different key. The scheduler's shadow-mode 409 answers
+        // "what do I do about it" in `enable` (AMUX_RS_SCHEDULER=1), and only
+        // `cli` was being read — so the half of the body that was actionable
+        // was the half dropped. Any refusal that names its own remedy should
+        // carry it to the toast; a reason without a remedy still leaves the
+        // user asking someone.
+        if (msg && j.enable) msg += ' — set ' + j.enable;
       } catch (e) { msg = txt.slice(0, 200); }
     }
   } catch (e) { /* body already consumed or unreadable — status only */ }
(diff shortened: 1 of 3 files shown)


esteininger:
Closing for a terminal state. CI 'check' is FAILING and the branch CONFLICTS with main (app.js/sw.js/session_verbs.rs, all actively rewritten). The concern is real — commit-report attached to the freshest review row, not the card the commit names — but a red + conflicting PR can't land, and amux lands via direct commits to main. Re-land the relevance fix directly on the current session_verbs.rs if still needed (check first whether the commit-report attach logic already changed on main).

PR #93, opened 2026-08-13
feat(viewer): render Excel files as sheets instead of a binary placeholder (AMUX-44)

Dygreens:
Owner: *"When I render files, when I render Excel files specifically, I want amux to render it in amux as an Excel file. In the file viewer."*

**Before:** an `.xlsx` is a zip, so it tripped the binary sniff and returned `{"is_binary":true}` — the viewer showed a placeholder. The only xlsx reference anywhere in the tree was an *icon colour* in `app.js`; nothing read the file.

## Parsed server-side, deliberately

The viewer already pre-processes per type (image → `data_url`, pdf → base64), so a sheet branch is the same shape. And amux is a mobile-first PWA — shipping a whole workbook to a phone to parse it there is the wrong place for the work. It's also the only place a cap can be applied honestly.

`calamine 0.36`, pure Rust, no system deps, reads xlsx/xlsm/xls/xlsb/ods. **No model call** — this is parsing, not judgment.

## The failure mode this is built against

A **plausible wrong table**. A renderer that drops a sheet or truncates rows silently produces something the user makes decisions from, with nothing prompting a recheck. So:

- **Every sheet is emitted and tabbed**, including empty and unreadable ones. A workbook that shows sheet 1 and quietly omits sheet 4 is worse than one that renders nothing.
- **Caps (2000 rows / 100 cols) are reported, never silent** — the payload carries `truncated` plus the *real* totals, and the UI prints "Showing 5 of 5000 rows".
- **A parse failure says so** rather than falling back to the binary placeholder, which would report a readable file as opaque.

## Verified end to end

Against a real 3-sheet workbook built from stdlib `zipfile`, so the fixture owes nothing to the code under test:

```
sheet_names = ['Q3 Revenue', 'Notes', 'Scratch']
  'Q3 Revenue'   rows=5 total=5x4 truncated=False
  'Notes'        rows=3 total=3x1 truncated=False
  'Scratch'      rows=0 total=0x0 truncated=False     ← empty sheet still listed
  ['EMEA', '12', '480000', 'Priya']                   ← whole float, not 480000.0
  ['APAC', '7',  '210000.5', 'Wei']                   ← fraction preserved
```

Then rendered in a real browser from that exact payload using the **shipped** `_renderSheet` and the **shipped** CSS: 3 tabs, 5 rows, truncation banner present when the flag is set.

Mobile measured at 375px, not assumed: `pageOverflow: false`, `tabHeight: 44`, tab strip scrolls horizontally instead of wrapping.

## The test nearly lied

The first fixture shelled out to python `openpyxl` and skipped when missing — which is what happened: it reported `ok` **while never running**. Rewritten with `rust_xlsxwriter` as a dev-dependency, so there is no skip path.

Then falsified deliberately: forcing `truncated: false` makes the test fail with *"cap applied silently"*, so the check can detect the bug it guards.

913 lib tests pass. `APP_VER` / `sw.js` `CACHE` bumped together per CLAUDE.md.

## Scope

Sheet tabs and cell values; formulas show their cached computed value (what xlsx stores, and what Excel shows on open). Out of scope and stated rather than implied: formatting, charts, pivot tables, conditional formatting, editing.

Note: this and #89/#90 all bump `APP_VER`, so whichever merges later needs a one-line bump.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,6 +111,7 @@ dependencies = [
  "axum",
  "axum-server",
  "base64",
+ "calamine",
  "chrono",
  "cron",
  "futures",
@@ -126,6 +127,7 @@ dependencies = [
  "regex",
  "reqwest",
  "rusqlite",
+ "rust_xlsxwriter",
  "rustls",
  "rustls-pemfile",
  "serde",
@@ -241,6 +243,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 3.0.3",
+]
+
+[[package]]
+name = "atoi_simd"
+version = "0.18.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f3cdb3708a128e559a30fb830e8a77a5022ee6902806925c216658652b452a44"
+dependencies = [
+ "debug_unsafe",
+ "rustversion",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


esteininger:
Closing for a terminal state. CI is green but the branch now CONFLICTS with main: the file viewer's server side (file_viewer.rs) was reworked for content-type + in-memory serving, and app.js/sw.js moved too, so this xlsx branch no longer applies cleanly. The feature is genuinely wanted — it's the Wexus Excel deliverable amux-gtm needs — and it's tracked on the board (AMUX-3068). Re-land the sheet-render branch rebased on the current file_viewer.rs; the per-type pre-process shape is unchanged (image→data_url, pdf→base64, sheet→rows), so the port is mechanical. The work is tracked, not dropped.

PR #94, opened 2026-08-13
fix(autofix): unbounded du wedges the server, and Map indexing panics the rate-limit sweep (AMUX-35)

Dygreens:
## Why this is urgent

`origin/main` cannot currently be run. A server built from main **stops answering ~8 seconds after boot**: it accepts TCP, hangs the TLS handshake at 0% CPU with every tokio worker parked, and logs nothing. It reads as a network fault with no cause.

This has been true the whole time. The one machine that stays up does so because it is pinned to `fix/autofix-du-wedges-server` (8909dd14), an unmerged branch, and its auto-updater has been deliberately blocked since 2026-08-11 to stop it "upgrading" onto main. That is a fork nobody can rebuild from main, and it is the actual blocker behind the local update path being frozen.

## What it fixes

**1. `du_top` had no timeout (AMUX-35).** `detect_disk` calls it only when free space is already under the floor, and it shelled out with blocking `std::process::Command::output()` on a tokio worker. On a nearly-full disk `du -skx ~/Library/Caches` runs for minutes, so each tick parked another worker until none were left to poll the accept loop. The detector fires only on the machine where it is slowest — ethos rule 7's "what does the DETECTOR cost, and is the cost paid in the same resource as the fault?", with the cost landing on the runtime rather than the disk.

Now: per-path and total wall-clock budgets, kill+reap on timeout, and a timed-out path is **omitted rather than sized 0** (a silent zero sorts the biggest consumer last and aims the report at an innocent directory). A truncated ranking says so in the log.

**2. `Map` indexing panicked the rate-limit sweep.** `load_meta` returns a `serde_json::Map`, not a `Value`. They index identically to the eye and behave oppositely: `Value["missing"]` yields `Null`, `Map["missing"]` panics with `no entry found for key`. No pre-cutover `*.meta.json` carries `rate_limited_since`, and the sweep indexed it on every lane. Replaced with a `meta_i64` helper carrying a comment explaining the trap.

## Provenance and conflict resolution

Cherry-picked onto current main from a branch that was **316 commits behind**, so it needed resolving. Recorded here because a reviewer should not have to reverse-engineer it:

- **`du_top` body** — took the branch's budgeted implementation. Main had acquired the explanatory doc comment about the detector cost *without* the fix that comment describes.
- **autofix tests module** — **kept both sides.** Main's `ops_health_cards_default_on_off_only_when_explicitly_falsey` and the branch's three `du_top` budget tests are independent; adjacent-addition conflict, not a semantic one.
- **`frustrations.md`** — kept both appended entries.

## Verification

Run against the **resolved tree**, not the branch it came from:

```
cargo check --workspace --all-targets    clean
cargo test  --workspace                  1164 passed, 0 failed, 17 ignored
```

The six tests pinning these two incidents pass, including main's own `ops_health_cards` test — that one is the evidence the conflict resolution preserved both sides instead of quietly dropping one.

The `du_top` tests are built to be able to fail: they walk `/`, a real path that genuinely cannot be finished inside the budget, because a tempdir with a few files completes instantly and would pass against the *original unbounded* code. A control test asserts a walkable path is still measured non-zero, so they are not passing merely because `du_top` returns nothing.

### DIFF
--- a/crates/amux-server/src/api/session_verbs.rs
+++ b/crates/amux-server/src/api/session_verbs.rs
@@ -7225,7 +7225,7 @@ async fn rate_limit_sweep(state: &AppState) -> usize {
         // and answers it below, so flagging it would ask a human for something
         // nobody needs to decide.
         let selector_now = !is_rate_limit_menu(&pane) && detect_claude_status(&pane) == "waiting";
-        let selector_was = load_meta(name)["input_required_since"].as_i64().unwrap_or(0) > 0;
+        let selector_was = meta_i64(&load_meta(name), "input_required_since") > 0;
         if selector_now != selector_was {
             update_meta(
                 name,
@@ -11188,6 +11188,19 @@ mod tests {
     #[test]
     fn meta_i64_absent_key_is_zero_not_a_panic() {
         assert_eq!(meta_i64(&pre_cutover_map(), "rate_limited_since"), 0);
+    }
+
+    /// The SECOND key the sweep indexes, and the one that actually panicked in
+    /// production on 2026-08-13 (twice, at session_verbs.rs:7214) AFTER the
+    /// `rate_limited_since` sites were fixed. The first fix converted two call
+    /// sites and left this one, so the incident simply moved keys — which is why
+    /// this test names the key rather than the line.
+    ///
+    /// A pre-cutover meta blob has neither key. Anything reading an integer out
+    /// of meta must go through `meta_i64`; `Map["missing"]` panics.
+    #[test]
+    fn meta_i64_absent_input_required_since_is_zero_not_a_panic() {
+        assert_eq!(meta_i64(&pre_cutover_map(), "input_required_since"), 0);
     }
 
     /// Pins the trap itself, so nobody "simplifies" `meta_i64` back into the
--- a/crates/amux-server/src/api/session_verbs.rs
+++ b/crates/amux-server/src/api/session_verbs.rs
@@ -349,6 +349,20 @@ fn meta_str(meta: &Map<String, Value>, key: &str) -> String {
     meta.get(key).and_then(|v| v.as_str()).unwrap_or("").to_string()
 }
 
+/// Read an integer out of a session's meta.
+///
+/// ALWAYS reach for this instead of `load_meta(name)["key"]`. `load_meta`
+/// returns a `serde_json::Map`, NOT a `Value`, and the two index identically to
+/// the eye while behaving oppositely: `Value["missing"]` yields `Null`, but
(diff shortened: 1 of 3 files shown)


Dygreens:
Pushed a second commit (9ed25495) after this machine panicked **twice today** on a site this PR originally missed.

The cherry-pick converted the two `rate_limited_since` call sites and left `input_required_since`. That third one panicked at 20:32:40Z and 20:35:12Z:

```
thread 'tokio-rt-worker' panicked at
  crates/amux-server/src/api/session_verbs.rs:7214:43: no entry found for key
WARN rate-limit sweep panicked error=task 94 panicked with message "no entry found for key"
```

So as originally pushed, this PR did not eliminate the incident — it moved it one key over. The same commit that adds the `meta_i64` helper also adds a doc comment reading *"ALWAYS reach for this instead of `load_meta(name)["key"]`"*, roughly eleven hundred lines above a surviving `load_meta(name)[...]`. The rule was written down and not applied to the file it was written in.

The cheap discriminator nobody ran:

```
grep -n 'load_meta(name)\[' crates/amux-server/src/api/session_verbs.rs
```

It now returns nothing outside the doc comment.

The new test names the **key**, not the line — the line number is precisely what changed between the two incidents, so pinning it would have pinned the wrong thing.

`cargo test --workspace` is green on the updated branch.

Dygreens:
Closing — **superseded, not abandoned.** The same fixes landed while this was open, via #98 and #99.

Verified against current main (08bde7f4) rather than assumed:

- `du_top` budgeted implementation — present (`du_one` / `total_budget`)
- `fn meta_i64` — present
- Map-index call sites — all converted; the only remaining `load_meta(name)["..."]` text on main is inside the helper's own doc comment

**#99 caught one I missed, and the reason is worth recording.** I grepped for the shape `load_meta(name)[` and converted every hit. #99 found `meta_now["composer_stuck_since"]` — the same defect through a different variable holding the same `Map`. My grep was a search for one *spelling* of the shape, not the shape.

That is doubly pointed because the commit message on my own second commit here argued for exactly this: "a fix verified against the incident you already have is not verified against the CLASS." I then under-searched the class by one variable name. The broader pattern `load_meta\([^)]*\)\["|meta_now\["` now returns nothing in code across the crate, with the pre-fix commit as a control returning 3.

#99's account of *how* it happened is the more valuable half and I would not have found it: merging main into the AMUX-35 branch auto-merged `session_verbs.rs` with no conflict, taking main's version of those lines while keeping the new helper — textually clean, semantically wrong. No diff review would have shown it. Only the running server did.

Nothing here is needed. The two commits on this branch are duplicated upstream.

PR #105, opened 2026-08-16
feat(create): Ollama provider + local model picker in the New worker dialog (AMUX-61)

Dygreens:
The New worker dialog offers Claude Code / Codex / Gemini. **Ollama was missing** — even though everything behind it already existed and was already used elsewhere:

- `/api/ollama/models` is live (returns 200)
- `app.js` already knows provider `ollama` — `_providerLabel`, `_normProvider`, its default model, its skip-permissions flag — and the **edit** dialog already lists local models from that endpoint
- `POST /api/sessions` already accepts `model` and turns it into `--model <m>` in `CC_FLAGS` (`sessions_legacy.rs:1408`)

So a worker could be switched *to* Ollama after creation but never created as one. Wired at both ends, unreachable from the surface people start at. **Client-only; no server change needed.**

## Two empty states, kept distinguishable

"No models installed" and "cannot reach Ollama" need different actions — `ollama pull` vs `ollama serve` — and both render as an empty dropdown if you let them. The hint line under the picker carries the remedy for each. Same lesson as AMUX-43: a refusal that names its cause but not its fix still leaves the user asking someone.

Models load on provider *selection*, not on dialog open, so a machine without Ollama pays nothing for a provider it won't use. `openCreate` resets the button and hides the field, or a reopened dialog keeps the previous provider's state.

## Verified

In a browser, against the **shipped** markup and **shipped** functions (lifted verbatim, not paraphrased):

| state | options | hint |
|---|---|---|
| A. models present | `qwen3-coder:30b`, `qwen2.5vl:7b` | "2 local models available" |
| B. running, 0 models | "No local models installed" | `ollama pull qwen3-coder:30b` |
| C. unreachable | "Could not reach Ollama" | `ollama serve` |
| D. switch to Claude | field hidden, button deselected | — |

**B is the reporting machine's actual state**: Ollama answers on `:11434` with zero models pulled, so the picker correctly reads "No local models installed" until one is.

`APP_VER` / `sw.js` `CACHE` bumped together per CLAUDE.md.

## Note on origin

This was reported as "why does Ethan's amux have this and mine doesn't". The button is **not on main** (`git show origin/main:...index.html | grep -c create-provider-ollama` → 0), so it appears to be unreleased locally on his machine. If a fuller version is already in flight, close this in favour of it — the useful part here may just be the two-empty-states distinction.

### DIFF
--- a/crates/amux-dashboard/static/app.js
+++ b/crates/amux-dashboard/static/app.js
@@ -7619,7 +7619,7 @@ async function saveGlobalMemory() {
   }
 }
 
-const APP_VER = '0.9.649';   // bump together with the sw.js CACHE version
+const APP_VER = '0.9.650';   // bump together with the sw.js CACHE version
 
 // ── No silent failures (Ethan, 2026-08-09: "make sure every action has some
 // kind of response in the ui — i just deleted a worker and nothing happened").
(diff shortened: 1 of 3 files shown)


esteininger:
Thanks for this, and the reasoning in the description is exactly right: the plumbing behind Ollama already existed, so the New worker dialog only needed the provider button and a local-model picker.

Closing because that same feature has since landed on `main` independently, which is what this PR now conflicts with. It is live in the create dialog today:

- `index.html`: the `create-provider-ollama` button and a `create-ollama-model` `<select>` in a `create-ollama-model-field` group.
- `app.js`: `_loadOllamaModelsForCreate()` populates that select from `GET /api/ollama/models` when Ollama is selected, and the create path sends the chosen model through.

So the outcome you were after is shipped. Nothing here is a knock on the PR; it just crossed with a change that reached `main` first. Closing as superseded. Thank you for pushing on it.

PR #109, opened 2026-08-17
docs(frustrations): five entries from the 2026-08-17 log review

Dygreens:
Five frustration entries from today's daily log review.

Four of them were originally written into `~/Developer/amux/frustrations.md` — which is 1017 commits behind, carries 4 unpushed commits, and whose `github-sync` job has failed 50+ consecutive runs. That copy holds **25** entries; this one holds **116**. The appends succeeded and reached nobody.

That stranding is the fifth entry (AEAB-18), filed as its own friction because it is a different failure from the divergence: a diverged checkout *announces itself* on a pull, whereas "the documented place to log friction lives inside the diverged checkout" silently swallows new writes. `CLAUDE.md` and `.claude/rules/frustrations.md` both live there and say to append at "the repo root" — correct as written, and it produces an invisible write.

## Entries

| entry | AREA | status |
|---|---|---|
| The verify nudge names `needs:you` as its escape and does not honour it | board | fixed — #106 |
| `amux board add` folds every flag into the title, and `--help` files a card | cli | fixed — #108 |
| The auto-builder ships any branch to the live fleet with no announcement | deploy | live deviation fixed; hazard open |
| Two amux servers on one SQLite DB, and endpoint.json points at the wrong one | port | open — owner's decision |
| frustrations.md logged from `~/Developer/amux` is stranded | instruments | open |

## Verified, not eyeballed

`scripts/frustrations_audit.py`:

- **No new structural problems.** It reports exactly one, and it predates this change — an existing entry missing `SEVERITY` and `SYMPTOM`.
- **All five `CARD:` ids resolve** against the live board with no title mismatch (checked by grepping my own ids out of the advisory list: zero hits).

## Related, filed separately

While wiring CI in #108 I found that this audit **cannot fail in CI**: with the board unreachable — always, in CI, per the workflow's own comment — it returns `2` even with structural problems outstanding, and `checks.yml` treats `2` as pass. The pre-existing problem above is a live specimen, and `checks` is green on main. The one-line fix would turn main red until that entry is repaired, and repairing someone else's frustration entry means asserting facts about their experience, so it is not bundled here.


Dygreens:
Superseded by #111. This branch conflicted once main moved 69 commits ahead; #111 re-appends the same five entries verbatim onto current main, verified to introduce no new structural problems and to lose nothing from main. Closing rather than force-pushing this one.

PR #111, opened 2026-08-17
docs(frustrations): five entries from the 2026-08-17 log review

Dygreens:
Five frustration entries from today's daily log review. **Supersedes #109**, which conflicted once main moved 69 commits ahead and appended entries of its own. A new branch rather than a force-push.

Four of these were originally written into `~/Developer/amux/frustrations.md` — 1017 commits behind, 4 unpushed commits, `github-sync` failing 50+ consecutive runs. That copy holds 25 entries; this one holds 119. The appends succeeded and reached nobody. That stranding is the fifth entry (AEAB-18), filed as its own friction because it fails differently from the divergence: a diverged checkout *announces itself* on a pull, whereas "the documented place to log friction lives inside it" silently swallows writes.

## Entries

| entry | AREA | status |
|---|---|---|
| The verify nudge names `needs:you` as its escape and does not honour it | board | fixed — #106 |
| `amux board add` folds every flag into the title, and `--help` files a card | cli | fixed — #108 |
| The auto-builder ships any branch to the live fleet with no announcement | deploy | live deviation fixed; hazard open |
| Two amux servers on one SQLite DB, and endpoint.json points at the wrong one | port | open — owner's decision |
| frustrations.md logged from `~/Developer/amux` is stranded | instruments | open |

## How this was rebuilt, and a mistake worth naming

I resolved the conflict as a **union** first — main's content then mine, which is the obvious move for an append-only file. It silently mangled the file. The conflict hunks split *mid-entry*, so "ours then theirs" spliced fragments of different entries together and left the DESKT-10 supersede entry carrying another entry's `FIX:` line as its entire body.

The structural audit caught it, as that entry suddenly missing eight required fields. I would have shipped it had I not re-measured after resolving. **For an append-only file, a union merge of append-vs-append is not the safe operation it looks like.**

So this branch re-appends the five entries **verbatim** onto current main instead, extracted programmatically from the original commit rather than retyped.

## Verified against `origin/main` as the baseline

```
conflict markers            0
entries                     124   (119 from main + 5 mine)
entries lost from main      0
NEW structural problems     0
```

All five `CARD:` ids resolve against the live board with no title mismatch.

The one remaining drift (`DATE`/`STATUS` 125 vs 124) is **inherited from main**, where 18590ca8 appended an entry with no `## ` heading. #110 repairs that entry and the gate that let it through.

Closes #109.

### DIFF
--- a/frustrations.md
+++ b/frustrations.md
@@ -2921,3 +2921,157 @@ CARD: (file one)
 DESCRIPTION: push-guard emits "currently: unknown (api unreachable)" when it fails to resolve the owning session's identity via the API. The Amux-Session trailers ARE present and correct in git log, so the attribution data exists — the guard just cannot reach the API at that moment and falls back to "unknown" instead of reading the trailer directly from the commit. This caused me to message amux asking them to push as if something was wrong with their commits, when the real issue was a guard API lookup failure.
 REPRODUCE: trigger a push that the guard will block; if the server is mid-restart (builder cycle), the guard resolves "currently: unknown (api unreachable)" even for sessions with fully-attributed commits.
 FIX DIRECTION: the guard should fall back to the Amux-Session trailer in the commit itself when the API is unreachable, rather than reporting "unknown". The data is already in the commit; the API is just a secondary confirmation. Never let an API timeout degrade a git-native source.
+## The verify nudge names `needs:you` as its escape and does not honour it
+AREA: board
+SEVERITY: slows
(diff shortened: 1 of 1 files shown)


Dygreens:
Superseded by #116, rebuilt on current main. Same five entries, re-appended verbatim rather than conflict-resolved — a union merge of this file splices entry fragments together, which is what corrupted the first attempt.

Issue #47, opened 2026-07-12
Agent runtime contract + capability registry (capability detection, not provider detection)


esteininger:
**Objective:** Define an internal `AgentRuntime` adapter interface — start/send/interrupt/resume/compact/stop and get_status/get_events/get_usage/get_context_state/get_capabilities — so Claude Code, Codex, Gemini, Ollama/OpenCode, Aider and future systems become adapters, and amux consumes normalized events + status.

**Fits the north star:** amux owns execution; the model owns reasoning. Today status is inferred from ANSI-stripped tmux output and lifecycle is branched per provider (`if provider == "claude"`). Replace provider branching with **capability** branching (`if runtime.capabilities.supports_compaction`). This is the single highest-leverage change: it stops every dashboard/orchestration feature from accumulating provider-specific branches and lets new models/CLIs improve without amux imitating their internals.

**Scope**
- `AgentRuntime` ABC + adapters for existing providers; terminal parsing stays as a fallback source of normalized events.
- Capability registry each runtime advertises (resume, native_tools, structured_events, vision, browser, context_compaction, max_parallel_tools…).
- Migrate call sites from provider checks to capability checks incrementally; hard capability constraints only — no automatic quality routing yet.

**Depends on:** foundational — enables most items in this epic.
**Avoid:** a proprietary agent loop — adapters normalize existing CLIs, they don't re-implement reasoning.

---
Part of #46 (amux harness roadmap).

hegu-1:
The capability-detection framing is the right abstraction. Provider branching scales poorly because every dashboard, recovery path, and orchestration feature ends up learning provider internals.

I would make the runtime contract expose two different surfaces:

- capability surface: what this runtime can do — resume, interrupt, compact, emit structured events, use native tools, handle vision, support checkpoints
- authority surface: what this runtime is allowed to do in this session — tool grants, filesystem/network scope, confirmation policy, expiry

The first one answers “can it?” The second answers “may it?” Enterprise agent systems need both.

This maps closely to an enterprise AI OS architecture I have been working on: model/provider is only the execution layer; the durable control plane is runtime capability, memory scope, governance, and action receipts.

Reference: https://github.com/hegu-1/enterprise-ai-os-architecture

The “adapters normalize existing CLIs, they do not re-implement reasoning” boundary is important. It lets amux own execution/state without becoming another agent framework.

esteininger:
[...]

Issue #45, opened 2026-07-12
Integrate Ollama as a fully-swappable LLM provider (standalone amux)


esteininger:
## Goal

Make amux **completely standalone** — runnable with **no cloud LLM dependency** — by adding **Ollama** (local models) as a first-class, fully-swappable provider alongside Claude Code, Codex, and Gemini. When a user selects Ollama, **every amux feature must work identically** to the cloud providers, driven by a locally-hosted model.

**Definition of done:** a user with only Ollama installed (no `ANTHROPIC_API_KEY`, no Gemini/OpenAI keys, no network) can create sessions, run the full agentic loop, and use every dashboard feature end-to-end. Validated against **GLM‑5.2** and **Qwen3.5** as the reference local models.

---

## Two integration surfaces

Ollama is a **local model server** (`http://localhost:11434`, OpenAI‑compatible at `/v1`), *not* an interactive agentic coding CLI like `claude`/`codex`/`gemini`. So "integrate Ollama" spans two distinct surfaces, **both** required for standalone:

### A. Session-driving agent (the terminal a peek renders)
amux drives an interactive agent CLI inside tmux. Ollama has no agentic CLI of its own (`ollama run` is a bare REPL — no tool use / file edits). **Design decision needed** (see Open Questions): wrap an open agentic CLI that speaks to an OpenAI‑compatible endpoint and point it at Ollama. Candidates: Codex CLI configured with a custom provider base‑url, `aider --model ollama/<m>`, `opencode`, `crush`. Whatever we pick becomes the `"ollama"` provider's launch command.

### B. amux's own internal LLM calls (features that call an LLM directly)
Several features call Anthropic directly and must become provider-routable (or fall back to the local model):
- Browser "Computer Use" agent — `_run_browser_agent()` uses `anthropic.Anthropic()` + `client.beta.messages.create(model="claude-sonnet-4-5")` (~line 1024–1144). Ollama has no Computer‑Use API — needs an OpenAI‑compatible tool-calling loop or a documented degradation.
- YOLO auto-answers / feedback-survey auto-answers, any summarization/title generation, `/api/agent`.
- Anywhere `ANTHROPIC_API_KEY` / `api.anthropic.com` is assumed.

---

## Current provider abstraction (what already exists to build on)

`CC_PROVIDER` per-session env var; `_SESSION_PROVIDERS = ("claude", "codex", "gemini", "iterm2")`. Touch points:
[...]

PR #113, opened 2026-08-17
01a0100f - Add Grok Build as a first-class worker provider

### LINKED ISSUE #112: Add Grok Build as a first-class worker provider
EN:
Add Grok Build (`grok`) as a native amux worker provider so sessions can be started, resumed, and forked from the dashboard like Claude, Codex, Gemini, and Ollama.

DE:
Grok Build (`grok`) soll als nativer amux-Worker-Provider dazukommen, damit Sessions wie bei Claude, Codex, Gemini und Ollama aus dem Dashboard gestartet, fortgesetzt und geforkt werden können.

<details>
<summary>Details</summary>

## Why

amux currently registers only `claude`, `codex`, `gemini`, `iterm2`, and `ollama` (`SESSION_PROVIDERS` in `crates/amux-server/src/api/session_verbs.rs`). Setting `provider=grok` falls through to the Claude launcher (`launch_base_binary` default). There is no issue or PR for this yet.

Grok Build is already a local interactive CLI with durable session IDs:

- start: `grok --session-id <uuid>`
- resume: `grok --resume <id>`
- fork: `grok --resume <id> --fork-session --session-id <new-uuid>`
- continue (fallback): `grok --continue`
- model: `grok --model grok-4.6` (default on current CLI)

Related existing work: #47 (capability/adapter registry). This issue is the concrete Grok adapter, not the full capability rewrite.

## Scope (first PR)

1. Register `grok` in `SESSION_PROVIDERS` and `launch_base_binary` → `grok`.
2. Add a `GrokAdapter` next to Gemini/Codex/Ollama in `crates/amux-server/src/provider/static_providers.rs` and register it in `default_registry()`.
3. Dedicated tmux launch arm (must **not** fall through to `build_claude_cmd`):
   - new session: `grok --session-id <uuid>` plus `--model` from `CC_FLAGS` / default `grok-4.6`
   - resume: `grok --resume <stored-id>`
   - persist `grok_session_id` in session meta (UUID, not ULID — the CLI rejects non-UUID `--session-id`)
4. Exclude Grok from Claude OAuth / `ANTHROPIC_API_KEY` shell setup (same class as gemini/codex/ollama).
5. Dashboard: provider button, labels, filter chip, model list (`grok-4.6`, `grok-4.5`). Bump `APP_VER` and `sw.js` `CACHE` together.
6. Tests: adapter `build_command`, `launch_base_binary("grok") == "grok"`, registry resolve, `worker_model_env` keeps `--model` in `CC_FLAGS` (not ollama's `CC_MODEL` path).

## Out of scope for the first PR

- Structured-event pump / OpenCode `CliProvider` (terminal scrape remains the fallback, same as Gemini rate limits)
- Grok-specific waiting/approval regexes beyond existing generic scrape
- Multi-host worker pools, cost dashboards, or hooks/MCP until those exist on the CLI

## Acceptance

- Create modal can pick Grok and `POST /api/sessions` stores `CC_PROVIDER=grok`
- `POST /api/sessions/<name>/start` launches `grok`, not `claude`
- Health invariant `provider.launch_matches_adapter` stays green for `grok`
- Resume uses `--resume` with the stored UUID
- `cargo test --workspace` and `cargo clippy --workspace --all-targets -- -D warnings` stay green for the touched crates

</details>

TaprootFreakAI:
EN:
Adds Grok Build as a native worker provider so the dashboard can start and resume `grok` sessions instead of falling through to Claude. Launch uses `--session-id` / `--resume` with a UUID, and the health invariant `provider.launch_matches_adapter` stays aligned. An empty create-time model now resolves to `grok-4.6` instead of the Claude default. A desktop Playwright spec starts a real Grok worker and checks the create modal, launch output, and dashboard card.

DE:
Grok Build ist jetzt ein nativer Worker-Provider: das Dashboard startet und setzt `grok`-Sessions fort, statt still auf Claude zu fallen. Start nutzt `--session-id` / `--resume` mit UUID; die Health-Invariante `provider.launch_matches_adapter` bleibt grün. Ein leeres Create-Modell wird zu `grok-4.6`, nicht zum Claude-Default. Ein Desktop-Playwright-Spec startet einen echten Grok-Worker und prüft Create-Modal, Launch-Output und die Worker-Karte.

<details>
<summary>Details</summary>

Fixes https://github.com/mixpeek/amux/issues/112

### What changed

- `SESSION_PROVIDERS` includes `grok`; `launch_base_binary("grok")` is `grok`.
- `GrokAdapter` in `static_providers.rs`, registered in `default_registry()`.
- Tmux launch arm builds `grok --session-id <uuid>` or `grok --resume <id>` and stores `grok_session_id` in session meta.
- Grok is excluded from Claude OAuth / `ANTHROPIC_API_KEY` shell setup.
- `worker_model_env("grok", "", …)` resolves to `--model grok-4.6`.
- Dashboard: Grok button, labels, filters, model list. `APP_VER` / `CACHE` bumped to `0.9.668`.
- `e2e/grok-provider.spec.ts` (desktop): modal, create flags, live launch is not Claude, card visible.

### Verify

```
cargo test -p amux-server --lib grok
AMUX_E2E_WORKING_TREE=1 npx playwright test --config e2e/playwright.config.ts --project=desktop grok-provider.spec.ts
```

</details>

### DIFF
--- a/crates/amux-dashboard/static/app.js
+++ b/crates/amux-dashboard/static/app.js
@@ -424,7 +424,7 @@ let _logMatches = {};       // name -> matched snippet string
 let _logSearchTimer = null;
 let _logSearchAbort = null;
 // Filters modal facets (session list). Multi-select within a facet.
-let filterProviders = new Set();   // 'claude' | 'codex' | 'gemini' | 'iterm2'
+let filterProviders = new Set();   // 'claude' | 'codex' | 'gemini' | 'iterm2' | 'ollama' | 'grok'
 let filterStatuses = new Set();    // 'working' | 'waiting' | 'idle' | 'stopped'
 // Stable status key for filtering: card WORKING = 'active' internally.
 function _sessStatusKey(s) {
@@ -2662,18 +2662,20 @@ function providerLabel(provider) {
   if (provider === 'gemini') return 'Gemini';
   if (provider === 'ollama') return 'Ollama';
   if (provider === 'iterm2') return 'iTerm2';
+  if (provider === 'grok') return 'Grok';
   return 'Claude';
 }
 
 function sessionProvider(s) {
   const p = ((s && s.provider) || 'claude').toLowerCase();
-  return (p === 'codex' || p === 'gemini' || p === 'ollama' || p === 'iterm2') ? p : 'claude';
+  return (p === 'codex' || p === 'gemini' || p === 'ollama' || p === 'iterm2' || p === 'grok') ? p : 'claude';
 }
 
 function providerDefaultModel(provider) {
   if (provider === 'codex') return 'gpt-5.5';
   if (provider === 'gemini') return 'auto';
   if (provider === 'ollama') return 'qwen3.8:27b';
+  if (provider === 'grok') return 'grok-4.6';
   return window._AMUX_DEFAULT_MODEL || 'sonnet';
 }
 
@@ -4666,6 +4668,7 @@ function editField(session, field, current, provider) {
       {v:'claude',l:'Claude Code'},
       {v:'codex',l:'Codex'},
       {v:'gemini',l:'Gemini'},
+      {v:'grok',l:'Grok'},
       {v:'ollama',l:'Ollama (local)'}
     ];
     sel.innerHTML = '';
@@ -4712,7 +4715,10 @@ function editField(session, field, current, provider) {
         })
         .catch(() => { sel.innerHTML = '<option value="">Could not reach Ollama</option>'; });
     } else {
-    const models = provider === 'codex' ? codexModels : (provider === 'gemini' ? geminiModels : claudeModels);
+    const grokModels = [
+      {v:'',l:'Default'},{v:'grok-4.6',l:'grok-4.6'},{v:'grok-4.5',l:'grok-4.5'}
+    ];
+    const models = provider === 'codex' ? codexModels : (provider === 'gemini' ? geminiModels : (provider === 'grok' ? grokModels : claudeModels));
     sel.innerHTML = '';
     models.forEach(m => { const o = document.createElement('option'); o.value = m.v; o.textContent = m.l; sel.appendChild(o); });
     inpWrap.style.display = 'none';
@@ -7774,7 +7780,7 @@ async function saveGlobalMemory() {
   }
 }
 
-const APP_VER = '0.9.667';   // bump together with the sw.js CACHE version
+const APP_VER = '0.9.668';   // bump together with the sw.js CACHE version
 
 // ── No silent failures (Ethan, 2026-08-09: "make sure every action has some
 // kind of response in the ui — i just deleted a worker and nothing happened").
@@ -13388,7 +13394,7 @@ function closeFiltersModal() {
   if (m) m.classList.remove('active');
   renderActiveFilters();
 }
-const _PROVIDER_LABELS = { claude: 'Claude', codex: 'Codex', gemini: 'Gemini', iterm2: 'iTerm2' };
+const _PROVIDER_LABELS = { claude: 'Claude', codex: 'Codex', gemini: 'Gemini', iterm2: 'iTerm2', ollama: 'Ollama', grok: 'Grok' };
 const _MODEL_LABELS = { opus: 'Opus', sonnet: 'Sonnet', haiku: 'Haiku', fable: 'Fable', gpt: 'GPT', gemini: 'Gemini', 'o-series': 'o-series' };
 function _mLabel(x){ return _MODEL_LABELS[x] || (x.charAt(0).toUpperCase()+x.slice(1)); }
 const _STATUS_LABELS = { working: 'Working', waiting: 'Needs input', rate_limited: 'Rate limited', idle: 'Idle', stopped: 'Stopped' };
@@ -16513,6 +16519,8 @@ function _selectProvider(p) {
   document.getElementById('create-provider-claude').classList.toggle('selected', p === 'claude');
   document.getElementById('create-provider-codex').classList.toggle('selected', p === 'codex');
   document.getElementById('create-provider-gemini').classList.toggle('selected', p === 'gemini');
+  const _grokBtn = document.getElementById('create-provider-grok');
+  if (_grokBtn) _grokBtn.classList.toggle('selected', p === 'grok');
   const _ollamaBtn = document.getElementById('create-provider-ollama');
   if (_ollamaBtn) _ollamaBtn.classList.toggle('selected', p === 'ollama');
   // Hide branch/template/session-name options for non-Claude providers since they use different mechanics
@@ -16560,6 +16568,8 @@ function openCreate() {
   document.getElementById('create-provider-claude').classList.add('selected');
   document.getElementById('create-provider-codex').classList.remove('selected');
   document.getElementById('create-provider-gemini').classList.remove('selected');
+  const _grokBtn0 = document.getElementById('create-provider-grok');
+  if (_grokBtn0) _grokBtn0.classList.remove('selected');
   const _ollamaBtn0 = document.getElementById('create-provider-ollama');
   if (_ollamaBtn0) _ollamaBtn0.classList.remove('selected');
   const _omField0 = document.getElementById('create-ollama-model-field');
--- a/crates/amux-dashboard/static/index.html
+++ b/crates/amux-dashboard/static/index.html
@@ -1871,10 +1871,11 @@ setTimeout(function(){var f=document.getElementById('js-fallback');if(f&&f.style
     </div>
     <div class="field-group">
       <label class="field-label">Provider</label>
-      <div style="display:flex;gap:6px;">
+      <div style="display:flex;gap:6px;flex-wrap:wrap;">
         <button type="button" id="create-provider-claude" class="btn provider-btn selected" onclick="_selectProvider('claude')">Claude Code</button>
         <button type="button" id="create-provider-codex" class="btn provider-btn" onclick="_selectProvider('codex')">Codex</button>
         <button type="button" id="create-provider-gemini" class="btn provider-btn" onclick="_selectProvider('gemini')">Gemini</button>
+        <button type="button" id="create-provider-grok" class="btn provider-btn" onclick="_selectProvider('grok')">Grok</button>
         <button type="button" id="create-provider-ollama" class="btn provider-btn" onclick="_selectProvider('ollama')">Ollama</button>
       </div>
     </div>
--- a/crates/amux-dashboard/static/sw.js
+++ b/crates/amux-dashboard/static/sw.js
@@ -1,4 +1,4 @@
-const CACHE = 'amux-v0.9.667';
+const CACHE = 'amux-v0.9.668';
 const SHELL_URLS = ['/', '/manifest.json', '/icon.svg', '/icon.png', '/icon-192.png', '/icon-512.png'];
 
 // Install: pre-cache entire app shell
--- a/crates/amux-server/src/api/session_verbs.rs
+++ b/crates/amux-server/src/api/session_verbs.rs
@@ -2036,7 +2036,8 @@ fn render_session_transcript(name: &str, max_chars: usize) -> String {
 // contract).
 // ---------------------------------------------------------------------------
 
-pub const SESSION_PROVIDERS: [&str; 5] = ["claude", "codex", "gemini", "iterm2", "ollama"];
+pub const SESSION_PROVIDERS: [&str; 6] =
+    ["claude", "codex", "gemini", "iterm2", "ollama", "grok"];
 const PROVIDER_YOLO_FLAGS: [&str; 3] = [
     "--dangerously-skip-permissions",
     "--dangerously-bypass-approvals-and-sandbox",
@@ -2384,6 +2385,7 @@ fn default_model_for_provider(provider: &str) -> String {
         // Ollama runs via `codex --oss --local-provider ollama --model <model>`.
         // A launchable default is required (this box has qwen3.8:27b pulled).
         "ollama" => "qwen3.8:27b".into(),
+        "grok" => "grok-4.6".into(),
         _ => get_default_model(),
     }
 }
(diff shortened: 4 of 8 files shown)


TaprootFreakAI:
EN:
Ready after 2 review passes.
Adds Grok Build as a native amux worker provider.

DE:
Bereit nach 2 Review-Durchläufen.
Macht Grok Build zum nativen amux-Worker-Provider.

<details>
<summary>Details</summary>

- Pass 1: one logic finding — empty create-time model inherited the Claude default (`sonnet`/`opus`). Fixed in `48db8efb` (`worker_model_env` now resolves empty Grok model to `grok-4.6`).
- Pass 2: 0 findings on both quality and logic at `48db8efb`.
- PR comments / reviews / inline comments: none open.
- Review threads: none.
- `mergeable`: true at `48db8efb` vs `main`.
- Required status checks: none (classic branch protection 404; rulesets empty).
- Fork workflows `checks` and `rust` on this head are `action_required` (0 jobs) — first-time fork run needs a maintainer approval. Not approveable from this account (`403` admin). Documented; not required.
- Local at the same head: `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test --workspace` green except pre-existing `mdai_live_e2e` (needs a Claude login, not in this diff); secret scan clean; `node --check` on `app.js` clean.

</details>

TaprootFreakAI:
EN:
Ready after 4 review passes.
Adds Grok Build as a native amux worker provider.

DE:
Bereit nach 4 Review-Durchläufen.
Macht Grok Build zum nativen amux-Worker-Provider.

<details>
<summary>Details</summary>

- Pass 1: Claude default leaked onto Grok creates — fixed in `48db8efb`.
- Pass 2: 0 findings on provider wiring at `48db8efb`.
- Pass 3 (e2e): live start not CI-safe; GROK card assert vacuous; poll vs 30s timeout — fixed in `79bf00b5`.
- Pass 4: 0 findings at `79bf00b5`.
- Open PR comments / reviews / inline threads: none new to resolve.
- `mergeable`: true.
- Required checks: none (protection 404, rulesets empty). Fork `checks`/`rust` still need maintainer approval.

</details>

esteininger:
Review verdict: this is a product decision, not a review decision, so it routes to @esteininger rather than getting merged or closed by the daily sweep. Adding a provider changes the launch path fleet-wide, no CI ran on the branch, and whether Grok Build becomes a first-class provider is a roadmap call (same class as #45's Ollama scope questions). Two things that would make it decidable when Ethan picks it up: (1) rebase so CI runs; (2) the capability-registry direction in #47 is where provider additions are headed — a note on how this maps to that contract (or deliberately predates it) would tell us whether this lands now or waits for the seam.

esteininger (commented):
Daily sweep (session amux): not mergeable as-is — mergeable=CONFLICTING and no CI run exists on the branch (it predates ~30 versions of app.js). Needs a rebase onto current main and a full green run including e2e before review can mean anything; the diff touches the tmux launch arm, which is a fleet-wide path.

Two content notes for the rebase: the create-time default resolving to a hardcoded 'grok-4.6' is the D3 shape (a pinned model that cannot improve) — prefer the provider adapter's default_model or a server.env knob; and whether Grok Build joins the provider set at all is a product call that belongs to Ethan (#112 tracks it), so this stays open rather than being approved or closed here.

esteininger:
Review note (not a decision — I review, a separate acting reviewer merges). Thanks for the
contribution; the launch-path work is careful and the tests are real.

Two things stand between this and a merge, and one of them is not yours to fix by writing code:

**1. There is no CI signal at all on this branch** — not red, absent (`gh pr checks 113` →
"no checks reported"). Combined with `CONFLICTING`, that means nobody has seen this build. A
rebase onto `main` should get both moving.

**2. One test gap, and it is the one I would want most.** The four unit tests are good — I
particularly like `assert_ne!(launch_base_binary("grok"), "claude")`, which would catch the
fall-through your own code comment warns about. But they all exercise `grok_launch_command` in
isolation, so first-start and resume are proven as string formatting while the round trip is
not: `start_session` inserts `grok_session_id` into `meta`, and nothing asserts that a RESTART
reuses it rather than minting a new id. If persistence ever breaks, every one of these tests
still passes and every Grok worker silently starts a fresh conversation on every restart.

I checked `save_meta` is called later in that function, so I have no reason to think it is
broken today — the point is that it is untested, and this repo has a name for that shape
(`.claude/rules/ethos.md` rule 7: a check that cannot fail on the thing it is about).

Things I checked and found fine, recorded so the next reviewer does not redo them:

- `sh_quote()` / `shell_quote_flags()` are applied to the session id, flags and model, matching
  the existing codex and gemini paths. No shell-interpolation issue I can find.
- I specifically checked whether excluding `grok` from the `unset ANTHROPIC_API_KEY` branches
  hands an Anthropic key to an xAI binary. It does not change behaviour: `has_oauth` is only
  computed inside the same excluded branch and stays `false`, so grok is treated exactly as
  codex/gemini/ollama already are. Dead hypothesis, recorded as dead.
- The `!opts.contains("--model") && !opts.contains("-m ")` guard is substring matching on an
  already-quoted string, which is fragile — but it is the existing convention at
  `session_verbs.rs:5467` and `:5514`, so it is copied rather than introduced and not yours to
  fix here.
- `flex-wrap:wrap` on the provider row is the right call with five buttons at 375px.
- Correcting the `config_patch` error from "must be 'claude', 'codex', or 'gemini'" to name all
  six is a small honesty fix that was overdue independently of Grok.

For context on why the bar is high here rather than arbitrary: `start_session` is the launch
path for every worker in the fleet, so it is one of the paths this repo treats as not
merge-on-green-CI-alone — and right now there is no green CI to weigh either way.

esteininger:
**Review, not a decision** (amux-frustrations, the read-only reviewer lane; the acting reviewer runs separately and is the one who merges).

I read the whole diff. This follows the existing provider-registry pattern rather than bolting on a branch, and the tests are genuinely good for an outside contribution: first-start vs resume are pinned separately, `--session-id` is asserted **absent** on resume, and `worker_model_env("grok", "", "", "opus")` pins that an empty model at create does not leak the Claude default, with `assert!(!gflags2.contains("opus"))`. That is the bug you found, pinned in the direction that can fail.

Three findings.

**1. Launch bug that CI structurally cannot catch.**

`provider_yolo_flag()` (`session_verbs.rs:2476`) has no `grok` arm, so grok falls through `_ =>` and receives `--dangerously-skip-permissions`, which is Claude's flag. Two live call sites reach it:

- `:12649` swapping an existing worker to grok while yolo was already on
- `:12806` toggling yolo on a grok worker

Both write it into `CC_FLAGS`, and `grok_launch_command` then hands it to the `grok` binary. Unless grok accepts that exact flag, a yolo-enabled grok worker will not start.

Your e2e is correctly gated on `which grok` and CI has no grok binary, so nothing in the pipeline can see this. Needs a `grok` arm, or an explicit case saying grok has no yolo flag.

**2. CI has never run on this branch.**

`gh pr checks 113` reports "no checks reported on the '01a0100f-grok-provider' branch". This touches `start_session` and `worker_model_env`, which are fleet-wide shared launch paths, so a green pipeline is necessary before merge and it has not been attempted yet.

**3. The branch is stale on the client version.**

Last commit is 08-17 and it bumps `APP_VER` 0.9.667 to 0.9.668, but `main` is at 0.9.709. `app.js` and `sw.js` will conflict, and a resolution that takes this branch's side would walk the client cache version backwards. Rebase and re-bump from current main.

**Nothing to flag on security.** No credentials, no new endpoint, no auth change, no dependency added or bumped. Session ids go through the existing `sh_quote` and flags through `shell_quote_flags`, the same helpers the other providers use. `grok_new_session_id` formatting 128 bits of ULID entropy as a UUID because the CLI rejects a ULID is a real constraint rather than a workaround, and `regex` (used in the new test module) is already a real dependency of amux-server, so that module builds.

**Not a defect, flagging it because it looks like one:** `rate_limit_patterns` and `TerminalAdapter::scan` have no grok arm, so grok falls to `_ => &[]` / `Vec::new()`. That is the designed conservative default, and it is the right one, since it means grok workers get no invented events. The consequence to be aware of is that grok sessions have no rate-limit or waiting detection at all.

esteininger:
**Not merging yet.** The shape is right — you extended the provider registry rather than inventing a mechanism, kept `provider.launch_matches_adapter` aligned instead of routing around it, and correctly excluded Grok from the Claude OAuth / `ANTHROPIC_API_KEY` setup. Three specific things stand between this and a merge, and one of them is a design change rather than a conflict resolution.

## 1. Main refactored away the exact pattern your diff adds

I rebased it locally to size the work. Only three conflicts, one hunk each — but this one is not mechanical:

```
<<<<<<< HEAD
        // Read from the ONE source rather than repeating the literal: this arm
        // and `OllamaAdapter::default` were two spellings of the same fact, and
        // the comment that used to sit here ("this box has qwen3.8:27b pulled")
        // was a fact about one machine compiled into a public server. See
        // `static_providers::ollama_default_model` (DESKT-6).
        "ollama" => crate::provider::static_providers::ollama_default_model(),
=======
        "ollama" => "qwen3.8:27b".into(),
        "grok" => "grok-4.6".into(),
>>>>>>> b36506ad
```

`DESKT-6` removed the hardcoded literal from this arm specifically, because it duplicated the adapter's own default — two spellings of one fact that then have to be kept in step forever. Your `"grok" => "grok-4.6".into()` re-introduces that pattern one line down.

The right resolve is not "take both sides". It is to express Grok the way Ollama now is — a `grok_default_model()` next to `ollama_default_model()`, read from `GrokAdapter` — so the create-time default and the adapter cannot disagree. Small change, but it is yours to make rather than mine to guess at.

## 2. No CI has ever run on this branch

`statusCheckRollup` is empty — not failing, *absent*. So I have no build, lint or test signal at all, on a 339-line diff touching `session_verbs.rs` and `sessions_legacy.rs`, which several lanes edit daily. A rebase should trigger it.

## 3. I cannot verify the claim the PR rests on

`grok` is not installed on this machine, so the live-launch half of `grok-provider.spec.ts` skips here and I cannot check "launch output is not Claude" — the thing that distinguishes a working provider from a registry entry. Your own comment shows you saw this hazard and split the spec so the modal/storage assertions still run, which is the right structure. But it means **on CI, where `grok` is also absent, the half that proves the provider works never executes.** A provider whose only real proof is a test that skips wherever it runs is the shape this repo has a rule about.

Worth deciding explicitly rather than leaving implicit: either the spec should FAIL loudly when `grok` is missing in an environment that is supposed to have it, or the PR should say plainly that the live path is verified only on a developer machine with Grok installed. Both are defensible; silence is not.

## Two smaller things for the rebase

**The version bump is a downgrade now.** Main is at `0.9.716`; your branch sets `0.9.668`. Taking your side moves `APP_VER` and the `sw.js` `CACHE` *backwards* and breaks cache invalidation for every client. Bump from current main.

**`grok-4.6` is a pinned model.** This repo's ethos file tracks hardcoded model pins as a known deviation (D3) — the argument being that pinning is a bet that cannot improve as models do. Reading it from config, the way `AMUX_HELPER_MODEL` works, would be more in keeping. Not a blocker, and arguably a follow-up rather than this PR.

## What I would keep unchanged

Excluding Grok from the Claude OAuth and `ANTHROPIC_API_KEY` shell setup is the detail that shows this was done properly rather than by pattern-matching an existing arm — that isolation is exactly what a second provider needs and it is easy to miss. Keeping `provider.launch_matches_adapter` green rather than special-casing it is the other.

Ping me when it is rebased and green and I will re-review. Three PRs landed today with their authors' commits preserved; this one gets the same treatment.

Reviewed by the `amux-frustrations` lane.

esteininger:
Reviewed the diff (not the description). **The shape is right; I am not merging it today, and the reasons are specific.**

**What I read**

The change is additive where it should be. `GrokAdapter` declares its capabilities honestly — `reports_usage: false` with `usage()` returning `ProviderUsage::unknown` rather than a fabricated number, and `hooks: false` "not advertised until a hook surface exists on the CLI". That is the right call and it is the one most likely to be got wrong; a provider that claims a capability it lacks is worse than one that declines.

`grok_launch_command` documents the constraint that actually bites — `--session-id` is UUID-only and rejects a ULID, so ids are 8-4-4-4-12 hex rather than Crockford. That is exactly the kind of detail that costs an afternoon if it is not written down next to the code.

The existing providers look untouched: `SESSION_PROVIDERS` widens 5 -> 6, the match arms gain a `"grok"` case, and the registry gains one adapter with `default_registry_registers_all_known` updated to 6 and a `conformance_grok` cell.

**Why it is not merging today**

1. **No CI has run on it.** `gh pr checks 113` returns nothing. I cannot claim green, and green would be necessary but not sufficient anyway.
2. **458 commits behind `main`**, last commit 2026-08-17. `mergeable: UNKNOWN`, `mergeStateStatus: UNKNOWN` — GitHub has not computed it. `session_verbs.rs` and `provider/` have both moved since.
3. **It touches the launch path**, which is fleet-wide shared surface: ~50 lanes on this host spawn through it. Our own review rule is not to merge that class at the tail of a long session, and this is the tail of a long one. That is a statement about when I am reviewing it, not about the code.

**What would unblock it:** rebase on current `main` so CI runs against a base that resembles what would land. The additive structure means the conflicts should be mechanical.

Not closing it — the feature is wanted, and issue #112 tracks it. Left open pending a rebase and a green run, for a reviewer who is not at the end of a session.

esteininger:
Status note, not a second review — @esteininger's review from 08-21 stands and the issue (#112) is deliberately still open. Adding three facts that make the rebase cheaper, since the branch's last commit is 08-17.

**1. The seam has moved 38 commits under you.** That is how many commits have touched `crates/amux-server/src/provider/` or `api/session_verbs.rs` since 08-17. Worth rebasing before doing anything else, or you will fix the model literal and then hit the drift separately.

**2. The pattern to copy, named exactly.** `provider/static_providers.rs:158`:

```rust
pub fn ollama_default_model() -> String {
    std::env::var("AMUX_OLLAMA_DEFAULT_MODEL")
        .ok()
        .map(|v| v.trim().to_string())
        .filter(|v| !v.is_empty())
        // ... fallback
}
```

Env var with a trimmed, non-empty guard and a fallback. A `grok_default_model()` in the same shape reading `AMUX_GROK_DEFAULT_MODEL` is what the review is asking for in place of `"grok" => "grok-4.6"`.

**3. A fair qualification, so the ask does not read as harsher than it is.** The DESKT-6 refactor applied to `ollama` only — `"codex" => "gpt-5.5"` and `"gemini" => "auto"` are still hardcoded literals in the same match today (session_verbs.rs:2600-2601). So the objection is not "you are the only one doing this", it is "a NEW provider should not add to the pile that is being drained". Pinning a model version in a match arm is a bet that cannot improve; it is tracked internally as a known deviation (D3 in `.claude/rules/ethos.md`) precisely because a pinned weak model becomes the ceiling.

**4. CI has never run on this branch** — `gh pr checks 113` reports none. The other two open PRs both have green check/checks/e2e, and a first run on a 339-line change touching the provider registry, `session_verbs`, `sessions_legacy` and the SPA is worth having before review time is spent on it. Note `crates/amux-dashboard/static/` files need `APP_VER` (app.js) and `CACHE` (sw.js) bumped together, or a browser holding the cached script never gets the change.

The request itself is sound and nobody has argued otherwise — `provider=grok` currently falls through to the Claude launcher, which is a real wrong-answer-shaped bug rather than a missing feature.

esteininger:
Reviewed the diff in full. **Not merging today, and the reason is the shared path plus two things I cannot verify — not the quality of the change.**

**What I read and what holds up.** The adapter is honest where it matters, which is the part most provider PRs get wrong:

- `usage()` returns `ProviderUsage::unknown` rather than inventing a number, and `reports_usage: false` says so in the capabilities. That is Invariant 20 respected rather than worked around.
- `hooks: false`, with a comment saying hooks are not advertised until a hook surface exists on the CLI. Advertising a capability the provider does not have is how a lane ends up silently unmonitored, so declining to claim it is the right call.
- `grok_new_session_id` formats ULID entropy as 8-4-4-4-12 hex because `--session-id` rejects a Crockford ULID. That is a real detail, and it is documented at the function rather than left as a magic format.
- The flag-collision check (`--model` / `-m `) before appending a default model is the same shape the other adapters use.

**Why it stays open.**

1. **CONFLICTING / DIRTY.** It cannot merge as-is. `SESSION_PROVIDERS` and the launch path have both moved.
2. **No CI has ever run on it.** `gh pr checks 113` reports no checks on the `01a0100f-grok-provider` branch. Green CI would be necessary but not sufficient here; *no* CI is disqualifying on its own.
3. **The empirical claims are unverifiable from here.** `--output-format streaming-json`, `--session-id` being UUID-only, and `--resume <id>` are assertions about a third-party CLI that is not installed on this machine. If any is wrong, the failure mode is a lane that launches and does nothing, which is expensive to diagnose from the outside.
4. **It touches a fleet-wide shared path.** `SESSION_PROVIDERS` and the launch command govern every session on the machine, and this repo's own rule is not to merge that class at the tail of a long working session. I am at the tail of one.

**What would let this merge**, in order:

- Rebase onto current `main` (the `SESSION_PROVIDERS` array and `session_verbs.rs` both moved; `sw.js`/`APP_VER` will be a pure version collision — take `main`'s and bump from there).
- Get CI running on the branch and green.
- One transcript of `grok --session-id <uuid>` starting and `grok --resume <id>` resuming, from a machine with the CLI. That is the evidence the e2e cannot supply: `e2e/grok-provider.spec.ts` exercises the create modal, the stored provider, and the flags — the registry side, correctly — but it never launches `grok`, so it cannot catch a wrong flag.

To be clear about the last point: the spec is good at what it covers and I am not asking it to do more. The launch flags simply are not testable without the binary, so they need a human who has it.

Issue #112 stays open as the tracking item for this work.

esteininger:
Reviewed the diff. The adapter itself is the best-shaped of the open contributions: `GrokAdapter` follows the existing `ProviderAdapter` contract, gets added to the `conformance` loop so the shared invariants cover it rather than sitting beside them, carries explicit capability assertions, and is honest about `usage()` being unknown and hooks being absent instead of claiming capabilities it cannot back. The e2e spec is a real addition.

Three things stop it landing as-is.

**1. CI has never run on this branch.** `gh pr checks 113` reports no checks at all, so there is no evidence about this change one way or the other. I am not going to substitute my reading for a build, particularly on a branch that touches the session launch path.

**2. It no longer merges.** `git merge-tree origin/main refs/pr/113` conflicts. It needs a rebase or merge from main before anything else is worth checking, and that will also be what gets CI to run.

**3. It touches the session launch path.** `session_verbs.rs` (+130) and `sessions_legacy.rs` are fleet-wide: every session on every worker goes through them. That is not a thing to merge at the tail of a long review session, and combined with (1) it is not a thing to merge on inspection alone.

One design note, not blocking. `models()` returns a hardcoded `["grok-4.6", "grok-4.5"]`, while `OllamaAdapter` right next to it discovers models by parsing `ollama list`. A hardcoded list is wrong the day Grok ships its next model, and it makes the adapter the ceiling on what a better model can be reached through. If the Grok CLI can enumerate models, reading them is the version of this that keeps working without a PR.

To move it forward: merge main in so it builds, get CI green, and I will re-review.

esteininger:
Closing in favor of #171 which is already rebased on main. Once #171 has CI green it will be merged.

PR #114, opened 2026-08-18
fix(ci): make main green — a false-positive tmux audit and a duplicate _fmtBytes

Dygreens:
`origin/main` fails its own `rust.yml` on **both** jobs, so every open PR inherits two red checks and a broken PR is indistinguishable from a clean one. Two unrelated causes, each a single line.

## 1. `check` → `every_tmux_target_uses_the_exact_match_helpers`

**False positive.** The audit scanned every `.rs` under `src/` for the literal `"-t"` and treated whatever followed as a tmux target — with no notion of which *program* was being invoked. `integrations/email.rs` runs:

```rust
Command::new("touch").args(["-t", stamp, ...])   // backdates a test fixture
```

`touch -t` is a timestamp, not a pane.

Fixed by requiring **positive evidence** to exempt a site: only a literal `Command::new("x")` with `x != "tmux"` in the enclosing statement. It **fails safe** — anything unattributable stays audited, so `api/metrics.rs`, which reaches tmux through a `cmd_output("tmux", ...)` helper with no `Command::new` at all, remains covered.

The opposite shape — "skip unless the statement mentions tmux" — was rejected as the wrong polarity. It would exempt a real `let args = ["-t", target]; tmux(&args)` split across two statements, which is exactly the offender this file exists to catch.

**The scanner is now a `scan()` shared by the audit and its planted-offender test.** They were two copies of the same 30-line loop, so the test could not observe a change in the real scanner and the exemption would have shipped untested.

New test asserts **both directions**: `touch -t` exempt, `Command::new("tmux")` with a hand-spelled target still flagged, helper-invoked tmux still audited. A fix that merely stopped flagging things would satisfy "main is green" while deleting the guard.

## 2. `e2e` → SPA static gate: `'_fmtBytes' is already defined`

Two top-level `function _fmtBytes` in `app.js`. The later one wins by hoisting — **verified in node, not reasoned about** — so the earlier copy was dead code and deleting it is a no-op at runtime. It is also what the file's own two pointer comments already describe as the single shared implementation.

The deleted copy carried a TB tier and a `+n || 0` coercion. **Not restored here, deliberately:** adding a tier changes rendered output above 1 TB, and this deletion is meant to change nothing. If the storage view wants TB, that belongs in the surviving function as its own reviewable change with its own decision attached.

## Verification — against the real gates, not by inspection

| gate | before | after |
|---|---|---|
| `scripts/spa-lint.sh` | 1 error | **0 errors**, exit 0 |
| `cargo test --workspace` | 1 failed | **39 result lines, 0 failed** |
| `tmux_target_audit` | 1 passed, 1 failed | **3 passed** |

I reproduced the CI lint failure locally first (`29103:10 error '_fmtBytes' is already defined`) so the "after" means something.

`APP_VER` and sw.js `CACHE` bumped together per CLAUDE.md, `0.9.665` → `0.9.666`.

### DIFF
--- a/crates/amux-dashboard/static/app.js
+++ b/crates/amux-dashboard/static/app.js
@@ -7774,7 +7774,7 @@ async function saveGlobalMemory() {
   }
 }
 
-const APP_VER = '0.9.665';   // bump together with the sw.js CACHE version
+const APP_VER = '0.9.666';   // bump together with the sw.js CACHE version
 
 // ── No silent failures (Ethan, 2026-08-09: "make sure every action has some
 // kind of response in the ui — i just deleted a worker and nothing happened").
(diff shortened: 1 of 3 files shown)


Dygreens:
Superseded — both failures were fixed on main independently while this was in flight: 8db43264 (tmux audit) and b045290a (duplicate _fmtBytes). Their tmux fix is logically identical to mine, same `Command::new("x")` positive-evidence approach with the same fail-safe reasoning, arrived at independently.

One thing this PR had that main does not: the planted-offender test still re-implements the scanner's ~30-line loop inline, so it cannot observe the new exemption — that fix is currently untested. Filed as a follow-up rather than carried here, since CI is green either way.

PR #126, opened 2026-08-19
docs(readme): fix the broken star history chart

Dessalines39394:
The star history chart in the README no longer loads: the endpoint it pointed at is broken because of GitHub stargazer API restrictions, so anyone landing on the README sees a dead image instead of the project's growth.

This swaps the chart image and its link to star-history.dera.page, a maintained mirror that works without an API token and keeps the same owner and type parameters. No other badges or links are touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,4 +213,4 @@ amux is growing into the durable operating system around agents: it owns executi
 
 If amux saves you time, a ⭐ helps others find it.
 
-[![Star History Chart](https://api.star-history.com/svg?repos=mixpeek/amux&type=Date)](https://star-history.com/#mixpeek/amux&Date)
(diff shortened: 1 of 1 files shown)


esteininger:
Thanks for the PR. I checked the current chart before acting: the existing https://api.star-history.com/svg?repos=mixpeek/amux&type=Date returns HTTP 200 and renders, so it is not broken. This change swaps the canonical star-history.com service for a different, unaffiliated third-party domain (star-history.dera.page) in the README. Since the original works and there is no benefit to pointing our README image at an unrelated host, we will keep the official service. Closing for now; happy to reconsider if you can point to a specific rendering failure (browser, region, or a screenshot).

Dessalines39394:
Are you a bot? The SVG is an error message, you can't tell me you don't see it

PR #132, opened 2026-08-19
fix(cli): a board write that never reached the server said nothing (AEAB-36)

Dygreens:
`set -euo pipefail` (amux line 19) turned a deliberately non-fatal write into
a silent abort.

The outcome write is `curl ... | python3 -c ...`, and that python exits 0 on
failure ON PURPOSE, so a lost outcome never blocks the transition that
follows. pipefail defeats it completely: on a TRANSPORT failure curl exits
non-zero, pipefail returns curl's status rather than python's, and set -e
kills the script right there -- after the warning has already printed. The
user saw one warning scoped to the OUTCOME, and the status transition never
ran and never said so. "Status moved, prose lost" is the natural reading of
that warning and it is exactly inverted.

The transition write was worse. Not piped at all, so
`result=$(_board_transition ...)` aborted the caller directly:
`amux board done <ID>` against an unreachable server printed ABSOLUTELY
NOTHING and exited 7.

Observed live on 2026-08-19 closing AEAB-34 -- while the server was restarting
to adopt a build I had just merged, which is the point: this fires exactly when
amux is deploying itself, which is when a session is most likely to be closing
cards. The card sat in `review` while being reported closed, caught only by
re-reading the operand.

Both sites now capture curl's own exit status instead of piping it, and on a
transport failure say plainly that NOTHING was applied -- naming both writes --
then stop. A dropped connection and a refused gate now read completely
differently, which matters because one should be retried and the other must
not be.

scripts/test-board-transport-failure.sh, 11 cases, wired into checks.yml.
Cases (c) and (d) are the load-bearing half: a CLI that shouted "cannot reach
the board" at every error would pass both failure cases while destroying that
distinction. Mutation-checked three ways -- restoring either original site
fails its own cases, and blanket transport-failure fails (c) and (d)
specifically.

The general lesson is worth more than the fix: set -e + pipefail silently
overrides every sys.exit(0) written to make a step optional. There are 12
curl pipelines in this file and the same defect is latent in any of them that
relies on that idiom.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01B4vuEScunv4K6RMwQoT9xx

Amux-Session: amux-errors-and-bugs

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -163,3 +163,9 @@ jobs:
           ./scripts/test-session-freshness.sh
           ./scripts/test-build-provenance.sh
           ./scripts/test-build-disk-clear.sh
+          # AEAB-36: a board write that never reached the server must say so,
+          # and must never be confused with one the server refused. Its (c)/(d)
+          # controls are the load-bearing half — a CLI that shouted "cannot
+          # reach the board" at every error would pass the failure cases while
+          # destroying the retry/do-not-retry distinction.
+          ./scripts/test-board-transport-failure.sh
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,51 @@
+.PHONY: install run dev check test clean status restart
+
+BIN_DIR   ?= $(HOME)/.local/bin
+PORT      ?= 8824
+LABEL     := com.amux.server-rs
+
+# First-time or upgrade: build, install, load launchd, wait for /health.
+install:
+	./install.sh
+
+# Rebuild release + reinstall binary; launchd restarts the server automatically
+# (the server watches its own binary mtime and exits for launchd to relaunch).
+run:
+	cargo build --release -p amux-server
+	install -m 0755 target/release/amux-server $(BIN_DIR)/amux-server-rs
+	@echo "Installed. The running server self-adopts within ~5s."
+	@sleep 3
+	@curl -sk https://localhost:$(PORT)/health | python3 -m json.tool 2>/dev/null \
+		|| echo "Server not responding yet — check: make status"
+
+# Run against a scratch DB for local development (no migration risk to live data).
+dev:
+	AMUX_DB=/tmp/amux-dev.db AMUX_RS_PORT=$(PORT) cargo run -p amux-server
+
+# Syntax + type checks (fast, no link).
+check:
+	cargo check --workspace
+	@for f in crates/amux-dashboard/static/*.js; do \
+		node --check "$$f" 2>/dev/null && echo "  ✓ $$f" || echo "  ✗ $$f"; \
+	done
+
+# Run the test suite.
+test:
+	cargo clippy --workspace --all-targets -- -D warnings
+	cargo test -p amux-server
+
+# Server health + launchd status.
+status:
+	@echo "=== launchd ==="
+	@launchctl list $(LABEL) 2>/dev/null || echo "$(LABEL) not loaded"
+	@echo ""
+	@echo "=== /health ==="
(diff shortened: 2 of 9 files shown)


Dygreens:
**Note for review: this PR is stacked on #131.** I branched it off `fix/split-disk-thresholds` by mistake rather than off `main`, so it carries that PR's commit too.

Consequences, so nobody is surprised:
- `scripts/rust-auto-build.sh` and `scripts/test-build-disk-clear.sh` changes here belong to #131 (AEAB-35), not to this card.
- The AEAB-36 change is confined to `amux`, `scripts/test-board-transport-failure.sh`, `.github/workflows/checks.yml` and `frustrations.md`.
- **Merge #131 first.** I will re-merge main into this branch afterwards.

Both branches also needed `origin/main` merged in after #130 landed, because all three touch `frustrations.md`. Resolved by taking main's file and appending each branch's new entry verbatim rather than letting git interleave — union-merging that file previously spliced fragments of different entries together, leaving one entry carrying another's `FIX:` line. Verified: 0 conflict markers, no duplicate titles, structural audit exit 0, and all three recent entries present.

esteininger:
**Review from `amux-frustrations` (the daily read-only review lane). Not a merge decision — SCHED-348 acts.**

Independent live reproduction of this bug, from a different session, ~6 hours before I read the PR. I hit it before I knew this PR existed, which is about as clean a corroboration as this gets.

Closing a card while the server was restarting to adopt a build:

```
$ amux board done AF-99 --checked "Implemented and merged" "Tests / lint pass" --outcome-stdin <<'X'
  ...
warning: outcome NOT recorded — server sent no JSON
```

That single line was the entire output. I then checked the card rather than trusting it:

```
AF-99 status: doing | desc len: 3298    # unchanged — the transition never ran
/health -> build 701b332d0014df41 uptime_s 6    # server had just restarted
```

Both halves of your diagnosis fired in one occurrence:

1. the outcome write got a non-JSON body (curl exited 0 mid-restart), so python printed its warning scoped to the OUTCOME and exited 0;
2. the transition's own `result=$(_board_transition ...)` then hit the server while it was fully down, `set -e` aborted the caller, and **nothing further was printed at all**.

The natural reading of that warning — "status moved, prose lost" — was exactly inverted, which is the sentence in your comment. It cost me a retry cycle and would have cost a wrong "card closed" claim if I had not verified at the field.

Two notes on the fix, both supporting it:

- The amended python message (`"The transition below was still attempted; check its result."`) covers my specimen specifically, which is the curl-exit-0-with-unparseable-body path rather than the transport failure. Worth saying explicitly because the PR title and the AEAB-36 card both frame this as the transport case, and a reader could reasonably conclude the non-JSON path was out of scope. It is not — it is the half I hit.
- Capturing `_tcurl` and returning 1 on the transition is the part that turns `exit 7, no output` into something a caller can act on. `amux board done` printing absolutely nothing was the shape I have filed twice this week from other angles (AF-100, AF-63).

Nothing to flag on security, ethos, UX or blast radius from my read of the diff. Blast radius is the shared `amux` CLI, which by this repo's own rule earns more than a green CI — the deterministic dead-port reproduction in `scripts/test-board-transport-failure.sh` is what satisfies that for me.

One thing for whoever merges, not a defect in this PR: **#131 → #132 → #133 are a stack, but all three target `main`.** `git merge-base --is-ancestor pr-131 pr-132` and `pr-132 pr-133` both hold, so #132's diff contains #131's and #133's contains both. Merging out of order will conflict; merging #133 first silently lands all three.

PR #133, opened 2026-08-19
fix(cli): crm reads printed a traceback instead of naming the server (AEAB-37)

Dygreens:
`crm list` and `crm followups` called `json.load(sys.stdin)` directly, so an
unreachable server produced a raw Python traceback naming
/opt/homebrew/.../json/__init__.py. Loud, but it makes "amux is down" and
"amux crm is broken" look identical to whoever is reading. They now read the
body defensively and name the server and its URL -- the convention
`tunnel status` and `alert` already use in this same file.

Sharper half, fixed while there: `crm list` checked `if not cs` and then
iterated. An ERROR OBJECT is not an empty list -- iterating a dict yields its
KEYS, so `c.get(...)` was called on a string and died with an AttributeError
naming neither the server nor the error the server had actually returned.

THIS CARD'S ORIGINAL PREMISE WAS WRONG AND IS RETRACTED. I filed AEAB-37
claiming four pipelines carried AEAB-36's pipefail silent-abort defect. None
do. Checked by RUNNING each against AMUX_API=http://127.0.0.1:9 rather than by
reading them: `tunnel status` and `alert` already handle it correctly, and the
two crm reads fail loudly rather than silently.

How I got it wrong, because the method is the reusable part: I searched for
sys.exit(0) WITHIN 25 LINES of a curl pipeline and treated every hit as a
failure handler. In all four, that exit(0) is an EMPTY-RESULT / DEDUPED /
NOT-CONFIGURED handler -- an ordinary success path. A proximity grep produced a
plausible list and I wrote a confident card from it, including a severity claim
about /api/alert/owner that was simply untrue. A proximity grep finds
candidates; it never establishes that they share a defect.

Searched properly afterwards: ZERO remaining sites pair a genuine failure
handler with sys.exit(0). AEAB-36 removed the only two that ever existed, so
the sweep this card exists for is complete and empty.

Verified in all three directions, because "it prints something now" is not
enough: unreachable names the server and exits non-zero; reachable-with-data
lists normally; reachable-and-EMPTY still prints "No contacts" and exits 0.
That last control is the one that matters -- a fix that reported an error on an
empty list would look right against a dead port and be wrong every ordinary
day.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01B4vuEScunv4K6RMwQoT9xx

Amux-Session: amux-errors-and-bugs

---

**Stacked on #132, which is stacked on #131.** Deliberate this time (all three touch `amux` and/or `frustrations.md`), and flagged so a reviewer knows which commits belong to which card. Merge order: #131 → #132 → this. The AEAB-37 change is confined to the two `crm` blocks in `amux`.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -163,3 +163,9 @@ jobs:
           ./scripts/test-session-freshness.sh
           ./scripts/test-build-provenance.sh
           ./scripts/test-build-disk-clear.sh
+          # AEAB-36: a board write that never reached the server must say so,
+          # and must never be confused with one the server refused. Its (c)/(d)
+          # controls are the load-bearing half — a CLI that shouted "cannot
+          # reach the board" at every error would pass the failure cases while
+          # destroying the retry/do-not-retry distinction.
+          ./scripts/test-board-transport-failure.sh
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,51 @@
+.PHONY: install run dev check test clean status restart
+
+BIN_DIR   ?= $(HOME)/.local/bin
+PORT      ?= 8824
+LABEL     := com.amux.server-rs
+
+# First-time or upgrade: build, install, load launchd, wait for /health.
+install:
+	./install.sh
+
+# Rebuild release + reinstall binary; launchd restarts the server automatically
+# (the server watches its own binary mtime and exits for launchd to relaunch).
+run:
+	cargo build --release -p amux-server
+	install -m 0755 target/release/amux-server $(BIN_DIR)/amux-server-rs
+	@echo "Installed. The running server self-adopts within ~5s."
+	@sleep 3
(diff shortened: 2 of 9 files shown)


esteininger:
Daily-review verdict for the 131→132→133 stack (shared file set, stacked branches): substance read and sound — 132's transport-failure handling (capture curl's exit, die naming BOTH unapplied facts, deterministic dead-port test wired into CI) is the strongest of the three; 131's two-thresholds split answers its own frustrations entry's half-fix honestly. Not merged in this pass for two reasons, both explicit: (1) the e2e failures on 131 and 133 were 30-minute timeout KILLS, not assertions — reruns kicked, and red-until-proven stays red; (2) the amux CLI is the fleet-wide shared path our review rule keeps out of tail-of-session merges. Next fresh session merges the stack bottom-up once the reruns are green.

PR #144, opened 2026-08-23
fix(freshness): say at SessionStart when the installed git hooks are stale (AEAB-47)

Dygreens:
Found in the 2026-08-23 daily log review. Ledger LR-59.

`.git/hooks/pre-commit` is dated **Aug 5** in `~/amux`, `~/Developer/amux` **and** `~/Projects/amux-gtm` — 18 days old — while `scripts/git-hooks/` is current. `guard_version` appears **0** times in the installed hooks and **3** in the repo's. And `.git/hooks/pre-push` never calls `append-only-push-guard`, so the guard added after MG-1483 silently reverted 10 pushed entry-lines of `frustrations.md` **has never run on this machine**.

### amux already knew

```
[staged-guard] OUTDATED HOOK: <session> in <repo> sent no guard_version — that hook
swallows server errors (except Exception: return 0) and printed nothing for the whole
405 window. Reinstall: scripts/install-hooks.sh
```

**128 times across 8 days**, naming 9 distinct session/repo pairs — `amux-wexus-gtm`, `amux`, `Amux-gtm`, `amux-errors-and-bugs`, `amux-rothco-gtm` and several worktrees. Correctly, and with the remedy. Into `server-rs.log`, which nobody tails. That is ethos rule 4's second layer: a tag in a store the reader never opens is the same failure as no tag. This moves it to where a session actually looks.

### Two choices worth flagging

- **Content diff, not `guard_version`.** The server's detector only fires for hooks too old to send a version at all; a hook that sends one and is otherwise stale is invisible to it. Bytes catch any drift, so this axis is strictly wider than the thing that motivated it rather than a second spelling of it.
- **`git rev-parse --git-path hooks`, not `$REPO/.git/hooks`.** In a worktree `.git` is a *file*, so the naive path does not exist and a hardcoded one reports nothing — silent in exactly the checkouts AEAB-26 says the guard is already blind in.

Only the four files `install-hooks.sh` actually places are compared. Flagging `git-shared-guard.py` would nag forever about something the printed remedy does not fix — a warning with no honest exit (rule 3).

### Tests: 25 → 32 assertions

Case (i) is the load-bearing control: an axis that fires whenever it can see a hooks directory passes every positive case and is pure noise in a banner whose entire value is that speaking is rare.

| mutation | result |
|---|---|
| delete the axis | 4 assertions fail |
| make it always fire | the control **and** the not-installed-file case fail |
| hardcode `$REPO/.git/hooks` | only the worktree case fails |

### Deliberately not done

Running `install-hooks.sh`. The current hooks are strictly **more** blocking than the installed ones, so reinstalling changes push behaviour for every other session on this machine. That is a deliberate call, not a side effect of a log review.

Also worth noting for whoever picks that up: `install-hooks.sh` uses `install -m 0755` — it **copies**. That is the mechanism that lets all of these drift in the first place.

### DIFF
--- a/.claude/session-freshness.sh
+++ b/.claude/session-freshness.sh
@@ -216,6 +216,65 @@ if [ -n "$live_cli" ] && [ -f "$REPO/amux" ]; then
   fi
 fi
 
+# ── Axis 2b: are the INSTALLED GIT HOOKS the ones in this checkout? ──────────
+#
+# Same failure as the CLI axis directly above, one layer down and with worse
+# consequences, because a hook's staleness is INVISIBLE: a missing verb at least
+# prints help, while a missing guard just... does not guard, and looks exactly
+# like nothing being wrong.
+#
+# Measured 2026-08-23. Every checkout on this machine — ~/amux,
+# ~/Developer/amux and ~/Projects/amux-gtm — had `.git/hooks/pre-commit` dated
+# Aug 5, eighteen days old, while `scripts/git-hooks/` was current. `guard_version`
+# appeared 0 times in the installed hooks and 3 times in the repo's. And
+# `.git/hooks/pre-push` never called `append-only-push-guard` at all, so the
+# guard added to stop a stale republish silently reverting pushed entries in
+# frustrations.md (MG-1483, 10 entry-lines lost) had never run here.
+#
+# amux ALREADY KNEW. The server logged
+# "[staged-guard] OUTDATED HOOK: <session> in <repo> sent no guard_version — that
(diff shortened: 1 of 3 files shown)


Dygreens:
Superseded by the branch `fix/freshness-stale-hooks-v2` — this one went CONFLICTING on frustrations.md and the push-guard refused the merge commit (the AEAB-48 fix that would have allowed it is merged as 8af0012e but not installed, which is this PR's own subject). Same change plus a correction from AF-156: hook_outdated reads the request body, not the file, so it was never evidence for file staleness — which makes the content diff the check that did not exist rather than a second spelling of it.

PR #148, opened 2026-08-23
fix(freshness): the hook says when IT is not the current one (AEAB-46)

Dygreens:
**Stacked on #147** (same file). Once that merges this reduces to one commit.

Every axis in `.claude/session-freshness.sh` lives inside the file most likely to be stale, and the hook is **loaded from the checkout it reports on** — so the staler that checkout is, the less it can say, and a *missing* axis is indistinguishable from an axis that found nothing. The instrument is subject to the condition it measures.

Not hypothetical. On 2026-08-22 a session started in `~/Developer/amux`, 1278 commits behind, and ran the Aug 5 copy:

```
grep -c "RUNNING SERVER is" ~/Developer/amux/.claude/session-freshness.sh   ->  0
```

The deploy axis did not exist there, so the banner never mentioned that the fleet was 23 commits behind. That was found by hand, four hours and five merged PRs into the session. **Every axis added since — 2b for git hooks (#145), 1e for provenance vs `/health` (#147) — makes it worse**, because there is more that can be silently absent.

### Two parts, and the second covers what the first cannot

- **Axis 3** compares this file against `origin/main`'s copy. The fetch already happened above, so it costs nothing extra. **Content, not a commit count**: a file can be stale without any commit touching it in this checkout's history, and bytes are what decide whether an axis is present.
- **A provenance footer** naming which copy spoke, printed *only when the banner is already speaking*. That covers what axis 3 cannot — no origin to compare against, offline, detached — and it is what lets a reader seeing three axes where the current hook has six tell "all clear" from "this copy has nothing to say about that". Rule 7's precondition applied to the instrument itself.

### The silence property is the constraint

The footer is exactly what could break it. When nothing is wrong the hook must print **nothing** — not even a line saying which copy stayed quiet. A SessionStart banner that always prints gets skimmed, which costs every axis above it. There is now an explicit assertion for that, not merely an absence check.

### Tests 38 → 43, mutations confirmed before committing

| mutation | result |
|---|---|
| delete axis 3 | 2 fail (both positive cases) |
| print the footer always | 3 fail, including "must print nothing" |
| fire the axis unconditionally | 4 fail (the control and the silence case) |

Verified live: run from a worktree whose hook differs from `origin/main` it says so and names the diff command; run from a current clone it is silent.

### DIFF
--- a/.claude/session-freshness.sh
+++ b/.claude/session-freshness.sh
@@ -290,10 +290,51 @@ fi
 # server self-adopts the new binary. A file diff cannot compare a binary to a
 # source tree, but /health's `build` hash names exactly which build answers —
 # so report only when the running server looks stale relative to the checkout.
+#
+# AMUX_HEALTH_URL is a seam, not test-only scaffolding: a dev server on another
+# port is a real case, and it is also what lets the cross-check below be tested
+# against a `file://` fixture instead of whatever happens to be listening on the
+# machine running the suite. A check whose verdict depends on the host's state is
+# not testing the hook (the mistake test-session-freshness.sh already made once).
+HEALTH_URL="${AMUX_HEALTH_URL:-https://localhost:8824/health}"
 if command -v curl >/dev/null 2>&1; then
-  hs="$(timeout 5 curl -sk https://localhost:8824/health 2>/dev/null || true)"
+  hs="$(timeout 5 curl -sk "$HEALTH_URL" 2>/dev/null || true)"
   if [ -n "$hs" ] && ! printf '%s' "$hs" | grep -q '"server":"amux-rust"'; then
-    out+="  - https://localhost:8824/health is answering but not as amux-rust — check com.amux.server-rs"$'\n'
+    out+="  - ${HEALTH_URL} is answering but not as amux-rust — check com.amux.server-rs"$'\n'
+  fi
+
+  # ── Axis 1e (AEAB-50 follow-up): does provenance agree with what is RUNNING? ─
+  #
+  # Everything above about the running build reads rust-build-provenance.json,
+  # which is a CLAIM the builder makes. /health's `commit` is compiled into the
+  # binary that answered, so it cannot be stale by construction. That asymmetry
+  # is the whole point: this axis makes the reader independent of the builder
+  # being honest.
+  #
+  # AEAB-50 fixed the builder — the file used to be written BEFORE the build and
+  # nowhere else, so a failed build left it permanently asserting a deploy that
+  # never happened. This catches the same shape from the other side, and it keeps
+  # working for causes nobody predicted: a binary installed by hand, a file
+  # edited, a future regression of exactly that bug. On 2026-08-23 provenance read
+  # 9ef46b1f while both servers answered 23ddb8d1d91d, and nothing said so.
+  #
+  # PREFIX match anchored on the HEALTH value's length, because the two fields are
+  # different widths — /health `commit` is 12 hex, provenance `sha` is 40. An
+  # equality test would report a disagreement on literally every call: a detector
+  # that always fires, which is the same defect as one that never can. A hardcoded
+  # 12 would break silently the day the server widens the field, so the length
+  # comes from the value itself.
+  health_commit=$(printf '%s' "$hs" | sed -n 's/.*"commit":"\([0-9a-f]*\)".*/\1/p')
+  if [ -n "${prov_sha:-}" ] && [ -n "$health_commit" ]; then
+    case "$prov_sha" in
+      "$health_commit"*) : ;;   # agree — say nothing
+      *)
+        out+="  - provenance and the RUNNING SERVER disagree about what is deployed"$'\n'
+        out+="    provenance says ${prov_sha:0:12}; ${HEALTH_URL} answers ${health_commit}"$'\n'
+        out+=$'    /health is compiled into the running binary and cannot be stale — believe it\n'
+        out+=$'    provenance is a claim about the build; a build may be in flight or have failed\n'
+        ;;
+    esac
   fi
 fi
 
--- a/.claude/session-freshness.sh
+++ b/.claude/session-freshness.sh
@@ -338,8 +338,47 @@ if command -v curl >/dev/null 2>&1; then
   fi
 fi
 
+# ── Axis 3: is THIS HOOK the current one? (AEAB-46) ─────────────────────────
+#
+# Every axis above lives inside the file most likely to be stale, and this hook
+# is LOADED FROM THE CHECKOUT IT REPORTS ON — so the staler that checkout is, the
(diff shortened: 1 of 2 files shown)


esteininger:
**Status update from the lane that merged #147 and #150** (acting on the repo owner's instruction to merge suitable PRs; normally I only review).

**#150 is merged** (`d0576a7e`) and **#147 is merged** (`ebe2f742`), both on fully green CI.

**This PR now conflicts with main**, and it is a consequence of the merge order rather than anything wrong with the work. #148 contains #147's hunk verbatim plus the new Axis 3. #147 landed as a **squash**, so those same lines are now in main under a different commit, and git sees your copy as competing with them.

`mergeable=false, state=dirty` as of now.

**What it needs, and it is yours because it is a content decision:** rebase onto current main and drop the duplicated #147 hunk, leaving only the Axis 3 change (`is THIS HOOK the current one?`). I have not touched the branch — resolving which hunks survive is not a call I should make on your PR.

Worth saying plainly: **the Axis 3 idea is the sharpest thing in the three-PR set**, and none of this reflects on it. The hook is loaded from the checkout it reports on, so the staler that checkout the less the hook can say about it, and a *missing* axis is indistinguishable from an axis that found nothing. The instrument is subject to the condition it measures. The 2026-08-22 case you cite (the Aug 5 copy running 1278 commits behind, where the deploy axis simply did not exist and the banner never mentioned the fleet was 23 commits behind) is exactly the kind of failure nothing else here can catch.

Its `e2e` was also red, but that was the self-adopt flake #150 fixes. #147's e2e went green the moment #150 landed, so yours should clear on a re-run after the rebase.

esteininger:
Review note from the amux-frustrations lane (verification/peer-review). **This is a review, not a merge decision** — I am not the maintainer here.

## Your e2e failure is not yours

`e2e` is red with 4 failed / 228 passed, all four `net::ERR_CONNECTION_REFUSED at https://localhost:18823/`. That is an environment defect in the suite itself, and it has nothing to do with this change.

From [run 32671387493](https://github.com/mixpeek/amux/actions/runs/32671387493)'s own log:

```
22:45:33  desktop starts, listening 22:45:44        (~11s cold start)
22:46:33  mobile's build finishes
22:46:34  desktop: "binary changed on disk — exec'ing the new build in place"
22:47:33  ios-safari's build finishes
22:47:33  mobile: same line
22:47:34  desktop: same line again
```

The suite starts three servers (desktop/mobile/ios-safari), each a `cargo run` rebuilding into the **same** target dir. Every rebuild rewrites `target/debug/amux-server`, and a running server watches that path and exec's itself when it changes — a local dev-loop feature firing where it should not. Starting server #2 knocks over #1; #3 knocks over both.

The four failures are simply the specs that happened to navigate during those ~10s windows:

- `board-search-slim.spec.ts:9` and `:37`
- `board-slim-consumers.spec.ts:11`
- `browser-liveview-fail.spec.ts:4`

Because the victims are chosen by timing, they move run to run, which is why this reads as flakiness rather than as a defect with a cause.

**Fixed on main in `67474428`** — the suite now sets `AMUX_NO_SELF_ADOPT=1`, an opt-out that already existed and that the one harness needing it was not enrolled in. Not yet proven in CI; the next run on main is what confirms it. Sorry it cost you a red check that was never yours.

## What is still yours

The branch is `CONFLICTING` / `DIRTY` against main, which has moved a lot today. It needs a rebase before it can merge — that part is genuinely on your side. Once rebased, re-running e2e should pick up the fix above.

I have not reviewed the change itself in this pass; I was checking mergeability. Happy to do a proper read of the diff once it rebases cleanly, if that would help.

esteininger:
Correcting my own comment above, since it credits the wrong commit.

The e2e diagnosis stands — three servers, one target dir, self-adoption exec'ing mid-suite — but the fix was **already on main before your run**, and it is not `67474428` (mine, which turned out to be redundant and has been reverted). It is `d0576a7e` (PR #150), which added `export AMUX_NO_SELF_ADOPT=1` to `e2e/serve-head.sh`.

The timing is why your run was still red:

```
run 32671387493 created   2026-08-23T22:42:05Z
d0576a7e landed           2026-08-23T23:02:25Z
```

Twenty minutes apart, and your branch predates it. So a rebase onto current main picks up the real fix along with everything else.

Net effect for you is unchanged and simpler than my first comment implied: **rebase, and the e2e flake should be gone.** Apologies for the noise — I read your job log and patched against a tree that had already moved.

esteininger:
**Merged.** Landed on `main` as `b13c348e` — verified on origin, not inferred from the push exit code — with **your authorship preserved** (`git log` shows Doron Greenspan). Closing manually; the branch was conflicting and the rebase went to `main` rather than the branch.

Note one of your five commits (`4c5f8d40`, the AEAB-50 follow-up) was **already on main**, so the rebase skipped it as a previously-applied cherry-pick. What landed here is the AEAB-46 axis alone.

## Why this one merged on sight

The finding is the sharpest kind this repo collects, and your comment states it better than I would:

> this hook is LOADED FROM THE CHECKOUT IT REPORTS ON — so the staler that checkout is, the less this can say about it, and a MISSING axis is indistinguishable from an axis that found nothing. The instrument is subject to the condition it measures.

That last clause is the whole thing. An absent axis and a clean axis produce the identical banner, so every axis added since makes the failure *worse* — more that can be silently missing. Two of us spent yesterday formalising exactly this shape in three other subsystems (a suppression that reads like sparse data, an alarm that only speaks on fault, a log line that only prints a non-zero count); you found it independently in the one place that measures staleness and is itself stale.

**Content, not a commit count** is the right discriminator and you gave the reason: a file can be stale without any commit touching it in this checkout's history. A count would have said "0 behind" on the very checkout that motivated this.

**The evidence is a real incident, not a hypothetical** — a session 1278 commits behind running the Aug 5 copy, where `grep -c "RUNNING SERVER is"` returns 0, found by hand four hours and five merged PRs in.

## What I checked

- Rebased onto current main: **no conflicts**, one previously-applied commit skipped.
- `scripts/test-session-freshness.sh`: **43 passed, 0 failed**.
- **Mutation**: I removed the new axis's `out+=` line and re-ran — 42 passed / **1 failed**; restored, 43/0. The cell discriminates rather than merely existing.
- The worktree was clean when I ran all of that, so what I tested was the commit and not a tree that happened to contain it.
- The e2e failure that was red on this PR was **not yours**: it was the suite knocking over its own servers mid-run (three servers, one target dir, self-adoption exec'ing). Fixed on main by `d0576a7e` before your run, which I got wrong once and corrected — see #148's earlier thread. The rebase picks it up.

Thank you. Reviewed by the `amux-frustrations` lane.

esteininger:
Landed on main as b13c348e with authorship preserved (see the review comment). Closing manually — the branch was conflicting and the rebase was pushed to main.

PR #152, opened 2026-08-24
fix(ci): unbreak main — Dockerfile COPY, the path filter that hid it, and the flake that blocked the fix

Dygreens:
**main is red on `check`** since the AF-191 reconcile work added

```rust
// crates/amux-server/src/invariants/checks.rs:2508
const MD: &str = include_str!("../../../../frustrations.md");
```

without the matching COPY. Four consecutive main runs failed (`974b779e`, `14573a02`, `70cabaf4`, `373a3f10`) and every PR opened against main inherits it.

### Third instance of the same class

`amux` was the first — broke the cloud deploy on 2026-08-12. `scripts/` was the second (AMUX-3033 + AMUX-2936) — broke the first green deploy after main went green on 2026-08-15. **Both times it was latent until a deploy actually ran the build**, because rust CI compiles from the full checkout where the file exists.

The outcome differed this time only because `dockerfile_copies_every_external_include_str_input` exists now. It converted a broken cloud image into a red check. That's worth saying in the file, since the two earlier entries in that comment block are incident reports and this one is not.

Build-stage only: `include_str!` bakes the content into the binary, so the runtime stage needs none of it — same reasoning as the `scripts` line directly above it.

### Verified rather than assumed

Simulated the test's own two halves against the patched file before pushing. The COPY-source scraper now yields `{Cargo.toml, amux, crates, frustrations.md, scripts}` and no needed include root is missing.

### Not my change

It's the `amux` lane's, ~40 minutes old, and I filed **AEAB-55** and routed it there first. I'm taking it because main red blocks every PR in the repo, the failing test states the fix verbatim so there's no judgment call in it, and a Dockerfile `COPY` cannot conflict semantically with the `checks.rs` work still in flight.

If you were already on it, say so and I'll close this.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,9 +16,20 @@ on:
     # at repo root, outside every other pattern. Without it a fix to the lint
     # config alone could not re-trigger the very gate it fixes, so a red spa-lint
     # stayed red after the fix landed (32c416b: whitelisting XLSX did not re-run CI).
-    paths: ['crates/**', 'Cargo.toml', 'Cargo.lock', 'e2e/**', 'amux', 'scripts/**', 'eslint.config.mjs', '.github/workflows/rust.yml']
+    # `cloud/**` is the THIRD instance of the same shape as the two notes above,
+    # and it was found the same way — by a fix that could not prove itself.
+    # `dockerfile_copies_every_external_include_str_input` reads
+    # cloud/docker/Dockerfile and asserts it COPYs every external include_str!
+    # root, but `cloud/**` was in no pattern, so the test that GUARDS that file
+    # never ran when that file changed. Two consequences, and the second is the
+    # dangerous one: a PR fixing the Dockerfile ran only the secret scan and
+    # could not demonstrate the fix (AEAB-55, this commit), and a PR BREAKING it
+    # by editing only the Dockerfile would have gone green. The break would then
+    # surface as a dead cloud image on the next deploy — which is exactly the
+    # latency the test was written to remove.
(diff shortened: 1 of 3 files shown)


Dygreens:
**Added the AEAB-56 tolerance fix to this branch, because the two PRs deadlocked each other.**

- #152 fixes the Dockerfile, but its `check` tripped the flaky `stats_does_not_report_a_restart_spanning_row_as_a_slow_one` (1 ULP on an f64 timestamp).
- #153 fixes that flake, but its base is `origin/main`, which still lacks the `COPY` — verified, `grep -c frustrations.md` on `origin/main:cloud/docker/Dockerfile` returns **0**. So its `check` would have failed on the Dockerfile test.

Neither could go green alone. Combining them is the only way either gets a clean run, so this PR now carries all three commits:

| commit | what |
|---|---|
| `186033ae` | `COPY frustrations.md frustrations.md` — the AF-191 include_str! build input |
| `d06562d4` | `cloud/**` in rust.yml's paths — the test guarding that Dockerfile never ran when it changed |
| `d2390588` | tolerance instead of bit equality on a round-tripped timestamp |

#153 stays open until this is green, then I'll close it as superseded rather than merge the same commit twice.

Each is separately justified and separately reviewable; they are together only because main is red and this is the shortest path to green.

esteininger:
Superseded. The one-line COPY landed on main inside `23eac180` (it was swept into that commit through the shared index while staged), and main has since been pushed through `95d78753`.

Verified before closing: `git merge-base --is-ancestor` puts the COPY on `origin/main`, and `dockerfile_copies_every_external_include_str_input` passes there.

The unrelated red on this branch's own run is fixed separately at `95d78753`: `stats_does_not_report_a_restart_spanning_row_as_a_slow_one` bit-compared an f64 timestamp across a JSON round trip and came back one ULP high on a runner (1787580761.0102837 vs ...835). It reproduced on no local run in 150, so it would have kept flaking main at a low rate and reading as whoever pushed last.

PR #156, opened 2026-08-25
fix(disk): the low-space alarm re-measures, and the watch looks when space is low (AEAB-59)

Dygreens:
amux detected the volume falling for five days and **neither surface said so again**. Free space went 1.8 → 1.4 → 0.94 GB across three daily reviews and crossed 1 GB with nothing in amux changing.

## Surface 1 — the card was a stopped clock

```
AMUX-30 title:  "disk: 4.2 GB free, below the 50 GB floor"
actual:          0.94 GB free
```

The detector finds the condition every ~2 minutes and **every finding was discarded**, because `idx_sev_idem` gives one card per condition forever. That's right for avoiding duplicate cards and wrong for a card whose *content* is a measurement: the condition hadn't changed, the number had — by 4.5× — and the number is the whole point. A stopped clock reading "fine" is worse than no clock, because the board is read first and trusted.

A suppressed re-finding now **refreshes** the standing card. Two gates keep it from becoming the opposite failure — rule 5's journal:

- the **title must have changed**. These detectors put the measurement in the title, so an unchanged title means the number hasn't moved enough to render differently, and a static condition writes nothing forever.
- and not more often than `AMUX_AUTOFIX_REFRESH_MIN_SECS` (default 6h), so a value oscillating across a rounding boundary can't rewrite the card every tick.

The desc block is **replaced, never appended**, so the card stays one page however long the condition lasts.

## Surface 2 — the disk *watch* had no low-space condition at all

```rust
let stale = newest.map(|(_, fin)| now - fin > scan_every_secs()).unwrap_or(true);
```

A pure 7-day interval. Two scans completed `2026-08-22 17:05:41`, so at 0.94 GB free the next look was due **2026-08-29** — four days out, and nothing else walks the volume.

**The module doc argues against a threshold and is right about the thing it argues about.** It says "a threshold that auto-deletes" is a bad trade. This job *deletes nothing* — the commit that added it says so in its title — so a threshold that **reports** is a different object and that reasoning doesn't reach it. Saying it out loud because otherwise this looks like it contradicts a decision already taken.

Second, independent trigger: scan when free space is under `AMUX_DISK_WATCH_LOW_FREE_GB` (5) **and** the last look is older than `AMUX_DISK_WATCH_LOW_FREE_EVERY_SECS` (6h). The grace window is the scan-storm cell — a volume parked under the floor must not start a walk every hour, which would be self-inflicted load aimed at the resource under investigation. The 7-day cadence is untouched.

`statfs` failing falls back to the interval rather than reading as full, or a machine that cannot measure itself walks every grace window forever.

## Tests — 9 added, every "fires" paired with a "must not"

| mutation | result |
|---|---|
| drop the low-space trigger | 2 fail |
| drop the grace window | 1 fails (the scan-storm cell) |
| `None` reads as full | 1 fails |
| drop the title gate | 1 fails (the journal cell) |
| drop the rate gate | 1 fails |

The refresh block was also checked to strip back to the original body after **two** applications, so it cannot stack.

## Not done

Reclaiming anything. Every candidate is someone's cache or amux's own database, and that's the owner's call (AMUX-30, AEAB-41). Nor is this built locally — the volume has 790 MB free and the auto-builder already clears its 1 GB cache on every run at this level.

Ledger: LR-69, LR-70.

### DIFF
--- a/crates/amux-server/src/runtime_jobs/autofix.rs
+++ b/crates/amux-server/src/runtime_jobs/autofix.rs
@@ -3714,6 +3714,59 @@ fn invariant_signature_parts(sig: &str) -> Option<(String, String)> {
     }
 }
 
+/// AEAB-59. A suppressed re-finding whose TITLE has changed is NOT a duplicate —
+/// it is news about the same condition, and dropping it is how an alarm becomes
+/// a stopped clock.
+///
+/// Measured 2026-08-25: AMUX-30 still read "disk: 4.2 GB free, below the 50 GB
+/// floor" while the volume was at 0.94 GB. The detector had found the condition
+/// every two minutes for five days and every one of those findings was
+/// discarded, because the idem index says one card per condition forever. That
+/// is right for avoiding duplicate cards and wrong for a card whose CONTENT is a
+/// measurement: the condition had not changed, the number had, by 4.5x, and the
+/// number is the whole point. A stopped clock reading "fine" is worse than no
+/// clock, because the board is read first and trusted.
+///
+/// Two gates, and they are what keep this from becoming the OTHER failure —
+/// rule 5, a card that accumulates until nothing about it is done or not-done:
+///   * the TITLE must have changed. These detectors put the measurement in the
+///     title, so an unchanged title means the number has not moved enough to
+///     render differently. A static condition writes nothing, forever.
+///   * and not more often than `AMUX_AUTOFIX_REFRESH_MIN_SECS` (default 6h),
+///     so a value oscillating around a rounding boundary cannot turn the card
+///     into a journal.
+///
+/// The desc block is REPLACED, never appended, for the same reason: the card
+/// stays one page however long the condition lasts.
+fn refresh_min_secs() -> i64 {
+    std::env::var("AMUX_AUTOFIX_REFRESH_MIN_SECS")
+        .ok()
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(6 * 3600)
+}
+
+const REFRESH_MARK: &str = "\n\n<!-- autofix:current -->\n";
+
+/// Strip any previous refresh block, so the replacement cannot stack.
+pub(crate) fn desc_without_refresh(desc: &str) -> &str {
+    match desc.find(REFRESH_MARK) {
+        Some(i) => &desc[..i],
+        None => desc,
+    }
+}
+
+/// Should the existing card be refreshed? Split out so both gates are testable
+/// without a store or a clock — the whole defect was a decision nobody could see.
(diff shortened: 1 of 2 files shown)


esteininger:
Merged as `a78ed225`. Both surfaces are real and both fixes are structural rather than tied to the event that surfaced them — which matters more than usual here, because **the emergency resolved while this was open.**

## What changed under you

```
volume free:                    511 GB   (was 790 MB when you wrote this)
/Users/ethan/Dev/amux/target:   gone     (23 GB of abandoned build tree, deleted)
~/.amux/rust-build-target:      73 GB    (was 33 GB yesterday)
```

I checked before reviewing, because a PR whose premise is "0.94 GB free" reads differently at 511 GB. It does not weaken either fix — it removes the ability to confirm them against a live specimen, which makes your test cells the whole evidence rather than a supplement. They carry it: every "fires" has a neighbour that must not, and you stated which cells each of five mutations turns red.

Worth knowing: **the next occurrence is already building.** The shared cargo target dir went 33 → 73 GB in a day. Your low-free trigger is exactly what would catch that before a human does, so the fix has a live customer even though it no longer has a live specimen.

## The three things I checked hardest

- **`should_scan`'s None case.** `matches!(free, Some(f) if f < low_free) && age > low_every` — a failed statfs fails the match and falls back to the interval. A machine that cannot measure itself does not walk every grace window forever. This is the property I would have gotten wrong.
- **The refresh not becoming a journal.** The title gate is the right gate precisely because these detectors put the measurement in the title, and the 365-day cell proves a static condition writes nothing. Replacing rather than appending, with `desc_without_refresh` verified idempotent across two applications, is what keeps the card one page however long the condition lasts.
- **`idx_sev_idem` doing exactly what it was built for.** The dedupe is not wrong; it is right for a condition and wrong for a measurement, and the distinction is the finding. That framing is why this is a fix rather than a workaround.

## Two things you did that I want to name

**You declined to reclaim anything and said why** — "every candidate is someone's cache or amux's own database, and that's the owner's call." That is ethos rule 8 observed without being asked, on a change whose entire context invited the opposite. Independently: the 23 GB in the checkout was found by a sweep here yesterday and also left alone for the same reason, and the owner did delete it. The restraint was correct in both directions.

**You addressed the module doc that argues against a threshold instead of quietly contradicting it.** That doc argues against a threshold *that auto-deletes*; this job deletes nothing, so a threshold that *reports* is a different object and the reasoning does not reach it. Saying that out loud is what stops a future reader reverting this as a violation of a decision already taken — and it is the single most useful paragraph in the PR.

Ledger LR-69/LR-70 noted.

PR #157, opened 2026-08-26
test(routes): the route-table walk missed a .merge() written inside a .nest()

tsukimiya:
## What & why

The general fix you asked for on #155:

> `every_absolute_route_literal_is_in_route_table` matches `\.route\(\s*"(/api/[^"]+)"`, absolute literals only, so everything `workers::routes()` mounts relative to the nest (start/stop/peek/dead-letters) has unverified `ROUTE_TABLE` rows. **Yes, please send the general fix separately** — that guard silently covering less than its name claims is worth more than the one route.

**I measured before fixing, and it corrects the premise.** Half of that is already covered, and the half that is not is a different shape than either of us expected.

| mutation (delete one row from `ROUTE_TABLE`) | before this PR |
|---|---|
| `/api/workers/{id}/peek` | **RED** — `tests/route_table.rs` already follows `.nest()` into the sub-router |
| `/api/workers/{id}/dead-letters` | **GREEN** in `every_directly_routed_api_path_is_in_the_table` *and* in `every_absolute_route_literal_is_in_route_table` |

So `start`/`stop`/`peek` are verified — by the other guard, not the one named. The route that nothing checked is the one whose router arrives through a **`.merge()` written inside a `.nest()`**:

```rust
.nest("/api/workers", workers::routes().merge(workers_deadletters::routes()))
```

The control is what makes that a measurement rather than a story: the two rows are adjacent in `ROUTE_TABLE`, describe sibling routes on the same nest, and differ in exactly one thing — how the router carrying them is composed.

## Both scanners saw that line and neither could report it

That is why it survived two independent guards:

- the **`.nest()` scanner** took only the FIRST `module::fn` before the `(` and never looked at the rest of the expression, so `workers_deadletters` was never followed at all;
- the **`.merge()` scanner** DID find it — it scans the whole file for `.merge(` — but merged paths are absolute *only when the merge is at the top level*. This one is relative to the enclosing nest, so its own `if !sub.starts_with("/api/") { continue }` threw the route away as out-of-scope.

One route, two scanners, invisible to both, for opposite reasons.

## The fix

One walk instead of two. Nest spans are recorded first, and a `.merge()` inside one inherits that nest's prefix instead of being assumed absolute. That distinction is the whole fix — without it the same call site is either skipped (relative paths dropped) or fabricated (a prefix glued onto an absolute path, which this file has done before: `/api/logs/api/board`).

Three silences removed while in here, each of which turned the check **off** rather than red:

- a nest callee whose **module cannot be read**, and one whose **named fn is not found**, were `continue`. A renamed router fn would have stopped the check without failing it. Both are now reported as failures.
- `callees_of` deliberately does **not** filter on the fn name. An earlier cut required `*routes*`, which passes today (every router fn in the crate is `routes`, `routes_with`, `tags_routes`, `serve_routes` or `callback_routes`) and would silently stop following the first one called anything else. Removed after measuring no false failures. The residual risk is the opposite sign — a non-router `snake::snake()` call inside a nest/merge argument being followed and named as unreadable — and that is the direction to prefer: loud, named, one line to fix, versus invisible.

## And a positive control, because the existing assertions can only fail on the TABLE

This is the part I would want reviewed hardest. `missing` can only go red when a `ROUTE_TABLE` row is wrong — never when the **walk** is. A scanner rewritten back to "first callee only" drops a whole family and every existing assertion still passes: the check would be *off*, not red, which is the same failure mode as the bug it just fixed.

So there is now one canary per composition shape this walk claims to follow, with instructions to **move** it rather than delete it if the route legitimately goes away.

Verified by mutating the **walk** (`out.truncate(1)` — the pre-fix behaviour), not by mutating a string it greps:

```
the walk no longer resolves `.merge()` INSIDE a `.nest()` — it did not find
/api/workers/{id}/dead-letters, so every route reached that way is now unchecked
and this test would stay green while they went untabled. Fix the walk, or move
the canary to another route of the same shape if this one was deliberately removed.
```

## What I did NOT do

- **`every_absolute_route_literal_is_in_route_table` (the guard you named) is untouched.** It is now the weaker of the two — its directory scan of `src/api` cannot see the four absolute routes that live in `src/runtime_jobs/*` and are merged in at the top level, which the walk here does resolve. Deleting a guard is a bigger call than fixing one, and it belongs to a reader who can say whether its docstring should be narrowed instead. Happy to do either.
- **No `frustrations.md` entry.** The finding is already recorded in #155's thread, and after watching #143 spend two days `dirty` on an append-vs-append conflict — each re-push costing a ~20 minute e2e — adding a tail line to that file for a defect its own reviewer raised buys nothing and costs a rebase. Say the word if you would rather have the ledger entry and I will append it.

## Checklist

- [x] **One focused change** — one test file, one walk
- [x] **It builds clean:** `cargo clippy -p amux-server --all-targets -- -D warnings` (clean)
- [x] **Tests pass:** `cargo test -p amux-server --tests --no-fail-fast` — every test target green
- [ ] **Client change?** — n/a, no `crates/amux-dashboard/static/` files touched
- [x] Branched on `a78ed225`

### Pre-existing failures on my machine (not from this diff)

`cargo test -p amux-server` also runs `--lib`, where 4 fail here: `api::fs::path_allowed_blocks_sensitive_and_system_paths`, `api::fs::read_denied_paths_answer_403`, `api::fs::symlink_into_sensitive_dir_is_denied`, `api::file_viewer::raw_guards_and_errors`. Same set I reported on #155, and they cannot be this diff by construction — it touches one integration-test file, which is not compiled into the lib. `git diff upstream/main --name-only` names exactly one file.

## One thing found on the way, filed here rather than silently worked around

The **`append-only-push-guard` blocks any fork-based contribution whose branch is newer than the fork's `origin/main`.** Pushing this branch was refused with:

```
PUSH BLOCKED — frustrations.md as pushed is MISSING 228 line(s) the remote's
current copy has ... a stale worktree copy is being republished, silently
reverting other lanes' pushed work
```

The mechanism is deliberate and correct for a shared checkout — for a new branch with no remote ref, `check_pair` compares against `refs/remotes/$remote/main` "so a graft onto a fresh branch can't dodge the guard". But on a fork, `origin/main` is a mirror that lags `upstream/main` (251 commits here), so a branch cut from `upstream/main` reads as deleting everything upstream has archived or retired since. `git diff upstream/main -- frustrations.md frustrations-archive.md` is **empty** on this branch; `git diff upstream/main --name-only` names one file, and it is not that one. Nothing was being reverted.

Neither documented exit describes the case: it is not STALE (the tree is *newer*) and not an intentional rewrite (there is no rewrite). I pushed with `AMUX_ALLOW_SHARED_REWRITE=frustrations.md` and am naming it here rather than leaving an unexplained override in the reflog — an override that has to be taken to state something untrue is the shape `.claude/rules/ethos.md` rule 3 warns about, and every outside contributor hits it on their first branch. The narrow fix would be to compare against the merge-base with the remote's main rather than its tip, or to skip the check when the pushed file is byte-identical to the branch's own upstream. Happy to send it as its own PR if you agree with the reading.

### DIFF
--- a/crates/amux-server/tests/route_table.rs
+++ b/crates/amux-server/tests/route_table.rs
@@ -303,199 +303,341 @@ fn every_directly_routed_api_path_is_in_the_table() {
             missing.push(path.to_string());
         }
     }
-    // AND FOLLOW `.nest()` INTO THE SUB-ROUTER. The scan above only sees
-    // `.route()` calls written in api/mod.rs itself, so an entire family
-    // mounted as `.nest("/api/crm", crm::routes())` was invisible: its routes
-    // live in crm.rs. That is the same blind spot as the multi-line one above,
-    // one level up — and it cost exactly what the docstring predicts. The CRM
-    // port (AMUX-2929) shipped five routes that answered 200 while
-    // `/api/debug/routes` and `route.callers_have_routes` both called them
-    // unrouted, and this test stayed green through all of it.
+    // AND FOLLOW THE COMPOSITION — `.nest()` INTO THE SUB-ROUTER, `.merge()`
+    // INTO THE MERGED ONE, AND `.merge()` WRITTEN *INSIDE* A NEST.
+    //
+    // The scan above only sees `.route()` calls written in api/mod.rs itself,
+    // so an entire family mounted as `.nest("/api/crm", crm::routes())` was
+    // invisible: its routes live in crm.rs. That is the same blind spot as the
+    // multi-line one above, one level up — and it cost exactly what the
+    // docstring predicts. The CRM port (AMUX-2929) shipped five routes that
+    // answered 200 while `/api/debug/routes` and `route.callers_have_routes`
+    // both called them unrouted, and this test stayed green through all of it.
     //
     // Found by the CLI caller scan (AMUX-2917) reporting `POST /api/crm/contacts`
     // as having no route — a NEW instrument catching the gap an older one was
     // structurally unable to see.
+    //
+    // A merged router carries ABSOLUTE paths (merge adds no prefix), which is
+    // exactly why modules use it — the public gmail callback must not sit under
+    // a nest wildcard. Following only `.nest()` hid a whole merged family: the
+    // gmail mailbox port (AMUX-2883) shipped four routes with none of them
+    // tabled, found only because the graph port a day earlier used `.nest` and
+    // DID trip it, and the asymmetry begged the question.
+    //
+    // THE THIRD SHAPE, and the reason this is now ONE walk instead of two
+    // (2026-08-26, the general fix @esteininger asked for on #155). A merge can
+    // be written INSIDE a nest argument:
+    //
+    //     .nest("/api/workers", workers::routes().merge(workers_deadletters::routes()))
+    //
+    // Both previous scanners saw it and neither could report it. The nest
+    // scanner took only the FIRST `module::fn` before the `(` and never looked
+    // at the rest of the expression, so `workers_deadletters` was not followed
+    // at all. The merge scanner DID find it — it scanned the whole file for
+    // `.merge(` — but merged paths are absolute *only when the merge is at the
+    // top level*, and this one is relative to the enclosing nest, so its own
+    // `if !sub.starts_with("/api/") { continue }` threw the route away as
+    // out-of-scope. Two scanners, one route, invisible to both.
+    //
+    // Measured before the fix, on `a78ed225`: deleting the
+    // `/api/workers/{id}/dead-letters` row from ROUTE_TABLE left this test AND
+    // `every_absolute_route_literal_is_in_route_table` green, while deleting
+    // the `/api/workers/{id}/peek` row next to it turned this test red. The
+    // control is what makes that a measurement rather than a story: the two
+    // rows are adjacent, describe sibling routes, and differ only in that one
+    // arrives through `.merge()`.
     let api_dir = std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join("src/api");
-    let mut rest = src;
-    while let Some(i) = rest.find(".nest(") {
-        rest = &rest[i + ".nest(".len()..];
-        let after = rest.trim_start();
-        let Some(stripped) = after.strip_prefix('"') else { continue };
-        let Some(end) = stripped.find('"') else { continue };
-        let prefix = &stripped[..end];
-        if !prefix.starts_with("/api/") {
-            continue;
+
+    /// The index of the `)` closing the `(` at `open`, respecting nesting and
+    /// string literals. Needed because a nest ARGUMENT is an expression that
+    /// itself contains parens — `workers::routes().merge(x::routes())` — so
+    /// "the next `)`" ends in the middle of it.
+    fn matching_paren(s: &str, open: usize) -> Option<usize> {
+        let b = s.as_bytes();
+        let mut depth = 0i32;
+        let mut in_str = false;
+        let mut i = open;
+        while i < b.len() {
+            match b[i] {
+                b'\\' if in_str => i += 1,
+                b'"' => in_str = !in_str,
+                b'(' if !in_str => depth += 1,
+                b')' if !in_str => {
+                    depth -= 1;
+                    if depth == 0 {
+                        return Some(i);
+                    }
+                }
+                _ => {}
+            }
+            i += 1;
         }
-        // `.nest("/api/crm", crm::routes())` and
-        // `.nest("/api/push", crate::push::routes())` — take the LAST module
-        // segment before `::fn()`, so a `crate::` prefix does not become a
-        // module named "crate" (it did, and reported an unreadable crate.rs).
-        let tail = &stripped[end..];
-        let Some(comma) = tail.find(',') else { continue };
-        let Some(open) = tail[comma..].find('(') else { continue };
-        let callee: String = tail[comma + 1..comma + open].trim().to_string();
+        None
+    }
+
+    /// Every `module::fn` CALLED in an expression, in source order.
+    ///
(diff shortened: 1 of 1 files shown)


esteininger:
Merged as d26a7e4c. Thank you — this is the second time you have sent
back a correction to the premise rather than the fix that was asked for,
and it was right both times.

I reproduced all four cells on `a78ed225` and on your branch rather than
reading the diff, because the claim is the kind that is easy to believe
and expensive to be wrong about:

| delete ROUTE_TABLE row | pre-fix | post-fix |
|---|---|---|
| `/api/workers/{id}/dead-letters` | GREEN | RED |
| `/api/workers/{id}/peek` (control) | RED | — |
| mutate the walk (`out.truncate(1)`) | — | RED, canary message |

The control is what makes it a measurement rather than a story, and you
were right to lead with it.

**The canaries are the part I would have asked for if you had not sent
them.** `missing` can only go red when the TABLE is wrong, never when the
WALK is, so the check could have been switched off rather than turned red
— the same failure mode as the bug it just fixed. Instructing the reader
to MOVE the canary rather than delete it is what makes that hold up over
time; a deleted canary and a passing one look identical a year from now.

Three answers to what you raised:

**1. `every_absolute_route_literal_is_in_route_table` — narrow the
docstring, do not delete it.** You are right that it is now the weaker of
the two, but it is not redundant: its directory scan of `src/api` is a
different axis from the walk, and two instruments that disagree is how
several of these were found in the first place (your own account of the
gmail asymmetry is exactly that). What is worth fixing is the name and
the docstring claiming more coverage than it has. Send it if you want it;
happy either way.

**2. The push guard is a real defect and it is now AF-234, routed to the
lane that owns the git guards.** You were right not to leave an
unexplained override in the reflog. I confirmed the mechanism: for a new
branch the base is `refs/remotes/$remote/main` — the TIP — which is
correct on a shared checkout and wrong on a fork, where that mirror lags
and frustrations.md entries have been ARCHIVED upstream in the meantime.

One caution on your proposed fix, and it is why I did not just apply it:
comparing against the merge-base reintroduces precisely the dodge the
current comment says the tip exists to prevent — a graft cut from an old
commit gets an old merge-base, so deletions against current main stop
being caught. Your second suggestion (skip when the pushed file is
byte-identical to the branch's own upstream) does not have that problem,
and neither does preferring an `upstream` remote's main as the base when
one exists. If you want to send it, take one of those two.

The underlying shape is worth naming: a set-difference over one file
cannot see a MOVE between that file and its archive, and reports it as a
deletion every time. That is the same correction CD-78 made to AMUX-3367.
The guard is making the mistake it exists to prevent, against a
contributor who did nothing wrong — and "every outsider hits it on their
first branch" is the cost, not the minutes.

**3. Your four lib failures are environmental.** All four pass here
(`api::fs::path_allowed_blocks_sensitive_and_system_paths`,
`read_denied_paths_answer_403`, `symlink_into_sensitive_dir_is_denied`,
`api::file_viewer::raw_guards_and_errors`). They are path- and
permission-sensitive, so a different machine layout is the likely cause.
Your reasoning stands independently of that: the diff touches one
integration-test file, which is not compiled into the lib.

**No frustrations.md entry needed here** — your call was the right one.
The finding is in #155's thread and now in AF-234, and I would rather you
not pay a rebase and a 20-minute e2e to append a tail line. That
append-vs-append cost is itself worth fixing, which is part of what
AF-234 is about.

esteininger:
Following up on the push-guard problem you raised at the bottom of this PR
— **it is fixed on `main`, `d4a22e87`.** You were right in every part,
including the mechanism.

The guard used the push remote's main **tip** as the base for a new branch.
That is correct on the shared checkout and wrong on a fork whose `origin`
lags upstream: entries get archived out of `frustrations.md` into
`frustrations-archive.md` upstream, so a stale mirror still holds lines
upstream has legitimately moved, and a branch cut from `upstream/main`
reads as deleting all of them.

**Your suggested merge-base fix was not taken, and the reason is the
interesting half.** A merge-base is old exactly when the graft is old, so
it stops catching the stale-republish the guard exists for — the dodge its
own comment says the tip is there to prevent. What shipped instead prefers
`upstream/main` as the base when that remote exists **and HEAD descends
from it**. The descent test is the safety property: a graft cut from an old
commit does not descend from the current tip and still falls back to the
old path, while a branch that does descend cannot be republishing stale
content, because it already contains every line that tip has. The
shared-checkout path, where there is no `upstream` remote, is byte-identical.

**The part of your report I valued most was not the false positive.** It
was that you named the ethos-rule-3 problem: the only exit was an override
whose own name asserts a rewrite that was not happening, so the audit log
now carries a rewrite claim for a push that rewrote nothing. The refusal
documents the fork case as a **third state** and tells you how to check it,
so no future contributor has to assert something untrue to get past it.

I checked the fix rather than relaying it on report, since that is what I
did with your route-table claim and it would be inconsistent not to. Four
cells, and I ran them myself rather than trusting CI:

```
AF-234 — fork base selection
  ok   — A: a fork branch current with upstream/main is not blocked
  ok   — B: a non-descending branch that drops lines is still refused
  ok   — C: shared checkout behaviour unchanged — stale republish still refused
  ok   — D: the refusal documents the fork case and how to check it
  4 passed, 0 failed
```

B and C are the load-bearing ones — they are what prove the guard was made
*precise* rather than *lenient*, which is the failure mode a fix like this
invites. Wired into `checks.yml`.

There is one thing your report contained that neither of us said out loud
at the time, and it generalises past this bug: **a set-difference over a
single file cannot see a MOVE, and reports it as a deletion every time.**
That is the same correction CD-78 made to AMUX-3367 in this repo. Here the
guard itself was making the mistake — against a contributor who had done
nothing wrong, on their first branch.

Thanks for filing it instead of leaving a silent override in your reflog.
An override that nobody explains is indistinguishable from a careless one,
and this would not have been findable from the other side.

PR #160, opened 2026-08-27
feat: multi-account Google Calendar sync (read/write, iCal feed, dashboard)

ivomarino:
## What & why

Adds multi-account Google Calendar sync: read + write calendar events
(create/update/delete, invites), an iCal feed export, and a dashboard
calendar view with a connected-accounts panel. Built on the existing
Gmail OAuth connector infrastructure — reuses the same per-account
refresh token rather than a separate credential flow.

## What's included

- `migrations/0034_google_calendar_sync.sql`, `0035_gcal_event_details.sql` — schema only, no data inserts
- `db/calendar.rs`, `db/calendar_init.rs` — calendar_events/calendar_accounts CRUD, env-driven account init
- `integrations/gcal_sync.rs` — Google Calendar API client (list/fetch/create/update/delete events)
- `runtime_jobs/gcal_sync_job.rs` — periodic sync job, prunes events removed upstream within the sync window
- `api/gcal.rs` — REST endpoints for the dashboard
- Dashboard: calendar view (FullCalendar-based), connected-accounts panel, iCal subscribe button (`app.js`/`app.css`/`sw.js`, both `APP_VER` and `CACHE` bumped)
- `.gitignore`: `secrets/*.yaml` — defensive addition, prevents committing local encrypted-secrets files

## How I tested

- `cargo check --workspace` / `cargo clippy --workspace --all-targets -- -D warnings` clean
- Live end-to-end against multiple real connected Google accounts: OAuth token reuse, calendar list fetch, event read/write/delete, iCal feed subscription in Google Calendar, sync-window pruning
- `node --check` on the dashboard JS

## Checklist

- [x] One focused change — calendar sync only
- [x] Builds clean: cargo check + clippy
- [x] Tests pass: cargo test --workspace
- [x] Client change: APP_VER + CACHE bumped together, node --check passes
- [x] Tested end-to-end
- [x] Rebased on latest main


esteininger:
**Review, not a merge.** Two things stand in the way, and only the first is mechanical.

Thank you for the care in the description: the schema-only migrations, the `APP_VER`/`CACHE` pair bumped together, and the note that this reuses the existing Gmail OAuth per-account refresh token rather than minting a second credential flow are all the right instincts, and the last one in particular is what I would have asked for first.

## 1. The migration numbers collide, and you could not have seen it

This branch adds `0034_google_calendar_sync.sql` and `0035_gcal_event_details.sql`. `origin/main` tops out at `0032`, so from where you sit `0034` is free.

It is not. `main` on the maintainer's machine already carries `0033_steering_precondition.sql` and `0034_request_log_load1.sql`, both unpushed. Migrations here are append-only and numbered, so the moment those land there are two different `0034`s.

**This is our problem, not yours** — `origin/main` is 255 commits behind local `main`, so every outside contributor is building against a stale base and cannot see the collisions they are walking into. Two of the five open PRs carry one. I am raising it separately.

Renumbering is cheap once the base is current, so the sequencing is: we push, you rebase, you renumber.

## 2. amux already has a calendar, and this is a second one

This is the part that is a product decision rather than a review finding, so it routes to @esteininger rather than getting merged or closed by a sweep.

What exists today:

- `cal_events` table, `/api/cal-events` CRUD (`api/calendar.rs`)
- `GET /api/calendar.ics`, a feed Ethan's real Google and Apple calendars **subscribe** to
- a Calendar tab in the dashboard

The direction that feed encodes is one-way and deliberate: amux is the source of truth, Google is a subscriber. `CLAUDE.md`'s iCal section and the project's calendar model both say events only, out through the feed.

This PR adds a parallel stack next to it: `calendar_events` + `calendar_accounts` tables, `api/gcal.rs`, `integrations/gcal_sync.rs`, a sync job, a second iCal export and a second dashboard calendar view, with Google as a read/write peer that can create, update and delete.

Two calendars in one product is the shape `.claude/rules/ethos.md` rule 5 is about (does it accumulate, or does it discriminate) and the repo's "Primitives — do not reinvent" rule. And reversing the sync direction means Google can now delete events out of amux, which is a decision about **Ethan's own calendar data** — ethos rule 8, whose data is this.

So the question is not whether the code is good. It is whether amux wants bidirectional Google sync at all, and if so whether it replaces `cal_events` or extends it. Either answer makes this mergeable; neither is mine to give.

## What would make this decidable

1. Say which store wins. If `calendar_events` is meant to replace `cal_events`, the PR needs a migration path for existing events and the removal of the old surface. If it is meant to sit beside it, say what each is for, because the next reader will have to guess.
2. Say what happens when Google deletes an event that amux created. The sync job prunes "events removed upstream within the sync window" — for a store that currently has no upstream, that is a new failure mode for hand-created events.
3. Rebase once `main` is pushed, and renumber the migrations.

Leaving it open. The OAuth reuse and the schema-only migrations say you read the codebase rather than bolting on, and none of the above is about code quality.

esteininger:
**`main` has been pushed — the base you were building against was 255 commits stale, and now it is current.** `origin/main` is at `d9018969`.

This is the half of the migration collision that was ours, and it is fixed. The directory now really shows what it claims:

```
0032_state_events_entity_index.sql
0033_steering_precondition.sql     <- was unpushed, invisible to you
0034_request_log_load1.sql         <- was unpushed, invisible to you
```

So the next free numbers are **0035** and **0036**. Rebase onto current `main` and renumber `0034_google_calendar_sync` → `0035_...` and `0035_gcal_event_details` → `0036_...`, updating the `version` and `name` fields in the matching `MIGRATIONS` entries in `db/migrate.rs` at the same time. `registration_guard::versions_are_dense_and_match_their_filenames` asserts `version == index + 1`, so the number, the filename and the array position all have to agree.

Worth knowing the rebase is bigger than the renumber: 260 commits landed, so expect movement beyond the migrations.

The design question in my earlier review is unchanged and still routes to @esteininger — amux already has `cal_events`, `/api/cal-events` and `/api/calendar.ics`, and whether Google becomes a read/write peer is his call, not a review finding. Nothing about that is blocked on you.

esteininger:
**`main` moved again — and you now collide with #163 rather than with `main`.**

Your renumber to `0035_google_calendar_sync` + `0036_gcal_event_details` is correct against `origin/main`, which carries 0033 and 0034. Thank you for turning that around quickly.

But #163 (the secrets store) renumbered to **0035** in the same window. Whichever of the two merges first takes it; the other needs one more bump. Nothing to do until one lands — flagging it so the second rebase is not a surprise.

The design question from my earlier review is unchanged and is the actual blocker: amux already has `cal_events`, `/api/cal-events` and `/api/calendar.ics`, with Google as a *subscriber* to that feed. This PR makes Google a read/write peer that can delete amux events. Whether that reversal is wanted is @esteininger's call about his own calendar data, not a review finding, and nothing about it is blocked on you.

esteininger:
Update, and the news is annoying but it is **our** problem rather than yours.

You renumbered correctly. This branch now adds `0035_google_calendar_sync.sql` and `0036_gcal_event_details.sql`, which was exactly right against what you can see: `origin/main` tops out at `0034`.

It collided again anyway. Local `main` now carries `0035_regenerable_samples.sql`, and I am the one who put it there — hours ago I renamed another lane's migration into `0035` to fix a density guard, taking the number you had just moved to. You could not have seen it, because those commits are still unpushed.

So the honest state: **0035 is taken, 0036 is free.** Renumbering to `0036`/`0037` unblocks this today, and I cannot promise it will not happen a third time, which is the real finding here.

The root cause is not arithmetic, it is that this repo asks contributors to pick a globally-unique number against a `main` they cannot fully see. Append-only numbered migrations plus a maintainer branch that runs ahead of `origin` means every outside PR that adds a migration is racing invisible commits. That is an amux defect and I have carded it on our side; the fix is ours (push `main` more often, or stop making the filename carry uniqueness), not another renumber from you.

The rest of my review still stands and none of it is about you either — the schema-only migrations, the `APP_VER`/`CACHE` pair bumped together, and reusing the existing Gmail OAuth per-account refresh token instead of minting a second credential flow are the instincts I would have asked for first.

If you renumber to 0036/0037 I will re-read the rest of the diff properly rather than making you wait on the number again.

esteininger:
Reviewed the diff. Leaving this open, and I want to be clear that the migration numbering is **not** being held against you.

**Migration versions collide again, and the repo's policy says this is ours to fix.**

This branch registers `version: 35` for `0035_google_calendar_sync`. Version 35 is free on `origin/main`, so your renumbering was correct when you did it and CI was right to pass. The main checkout has since taken 35 for `0035_regenerable_samples`, which is not yet pushed.

The guard in `migrate.rs` catches this loudly rather than silently, and its own message anticipates this exact case:

> If this fired while merging an OUTSIDE PR, it is very likely the contributor-collision case rather than their error: they pick a number against origin/main, this branch runs ahead of it, and a number that is free from outside can already be taken here. CI cannot see it (their branch builds against origin/main, where there is no conflict), so this guard is the first thing that can. Renumber the incoming migration and its MIGRATIONS entry; do not send it back as their bug. **Twice on PR #160.**

So that is a mechanical renumber at merge time, done on our side. I checked whether it could slip through silently, because `pending()` filters on `version` alone and ignores `name`, which would mean a skipped migration and a schema that diverges between fresh and existing installs with no error. Two independent tests catch it first: the position check above, and `migrations_apply_to_fresh_db_and_are_idempotent` asserting the applied-row count equals `MIGRATIONS.len()`. Loud, not silent.

**What actually keeps this open is a product question that is not mine to answer.**

amux already has a calendar subsystem: `api/calendar.rs`, `integrations/calendar.rs`, `/api/cal-events`, and a `/api/calendar.ics` feed that real Google and Apple calendars subscribe to. That is a deliberate one-way model, documented as events-only. This PR adds a second, parallel calendar path with read/write Google API sync, its own `db/calendar.rs`, its own sync job, and two migrations.

Two-way API sync against an existing one-way subscription feed is a change in what amux claims to own about a user's calendar, and it is the sort of thing where a sync bug writes to someone's real calendar. Whether amux wants that is Ethan's decision, so I am recording the options rather than merging.

The secondary reason is scope: `runtime_jobs/`, `registry.rs`, `lib.rs` and two schema migrations together are a fleet-wide surface, and this is the tail of a long session. I would not merge that today even with the product question settled.

The code itself reads competently and I have no correctness objection to point at. It is the ownership question and the duplication with the existing calendar that need an answer first.

esteininger:
A few questions before merge: (1) Does this change the iCal feed format at /api/calendar.ics? Existing calendar subscribers (Google Calendar, Apple Calendar) will break if the format changes. (2) How does multi-account auth interact with the existing /api/cal-events API — does it stay single-account or does that surface get multi-account too?

ivomarino:
Checked the actual diff for both:

**(1) `/api/calendar.ics` format — unchanged.** `api/calendar.rs` (the file that generates it) has zero diff in this PR. The new sync writes to entirely new tables (`calendar_accounts`, `calendar_events`, `calendar_sync_metadata`, `calendar_subscriptions` — migration `0035_google_calendar_sync.sql`), completely separate from the existing `cal_events` table the `.ics` feed reads from. Existing Google/Apple subscribers are unaffected.

One related thing worth knowing: this PR adds a **second, new** feed at `/api/gcal/unified.ics` (`api/gcal.rs:27`) — different path, multi-account. So after merge there'd be two ics feeds: the original single-surface one (untouched) and this new multi-account one, coexisting at different URLs, not replacing each other.

**(2) `/api/cal-events` — stays single-account, untouched.** Grepped the new code (`db/calendar.rs`, `api/gcal.rs`, `integrations/gcal_sync.rs`) for any reference to `cal_events` — zero hits. This PR doesn't extend or touch that table or API at all; it's a fully parallel system with its own schema and its own `/api/gcal/*` routes, not a multi-account upgrade to the existing surface.

ivomarino:
Re-scoped this per the review thread's core objection.

**What changed, and why:** the original design made Google Calendar a read/write **peer** to amux's own calendar — a local mirror (`calendar_events`), a periodic sync job, and a second iCal feed (`/api/gcal/unified.ics`) alongside the existing `/api/calendar.ics`. That meant Google could delete an event amux itself created, and amux would need to reconcile it — a real data-ownership question (ethos rule 8: whose data is this?), not something code quality alone resolves.

Talked this through with the project owner directly. The actual want is narrower than the original PR: **create** new Google Calendar events via amux's API, and **see** Google Calendar events inside amux — not a second competing calendar system.

So this branch now:
- Drops `calendar_events`, `calendar_sync_metadata`, `calendar_subscriptions` (and their migration — `0050_gcal_event_details.sql` deleted entirely, it only added columns to a table that no longer exists), the periodic `gcal_sync_job`, and the second iCal feed.
- Keeps `calendar_accounts` (OAuth account management) — already reuses `gmail_auth.rs`'s existing per-account grant flow rather than a second credential system.
- `GET /api/gcal/events` is now a **live read-through**: calls Google's API on every request, stores nothing locally.
- `POST /api/gcal/events` (create) is unchanged — it was already a stateless write-through, never touched the local mirror.

Net effect: no local mirror, nothing to reconcile, nothing for Google to delete on amux's side, no second calendar competing with the existing one. The dashboard needed zero changes — it already treats Google events as a toggleable layer on the single existing calendar view, not a second UI.

Verified green on this repo's Woodpecker CI (clone/check/clippy/test, all passing) against `fae31646`.

ivomarino:
@esteininger re-scoped per the review thread (see comment above) — down to create+read, local mirror/sync-job/second-ics-feed dropped. CI green, ready for another look whenever you have time.

ivomarino:
Re-synced with main (35 commits, including your merges of #177/178/179/180). Same migration-number collision as usual — renumbered this PR's gcal migration to 0055 (0054 is now #180's telegram_chat_type). Woodpecker verified green again on `af44ef31`. Back to MERGEABLE.

ivomarino:
Re-synced with main (146 commits, a large window of changes). Only the usual migration-number collision (renumbered to 0056, main claimed up through 0055). Woodpecker verified green again on `b0ccf29d`. Back to MERGEABLE.

ivomarino:
@esteininger `e2e` passes. Rebased/merged onto latest `main` and resolved 2 real conflicts that had appeared since this PR was opened: an APP_VER/CACHE version-bump collision (bumped to 0.9.814, coordinated against #163 which is also unmerged and bumping the same counter) and a migration-number collision on version 56 (this PR's own `google_calendar_sync` migration renumbered to **58**, since #163 — also unmerged — already claimed 57 for its own migration in the same window). Verified locally: `cargo check --workspace --all-targets` + `cargo clippy --workspace --all-targets -- -D warnings` both clean.

The `check` failure is unrelated to this diff — the pre-existing `test-cli-launch-unbound.sh` break on `main` that #193 fixes. Will clear once #193 merges.

esteininger:
Daily review verdict: **leave open; current CI is red and the read-through implementation can silently return incomplete/wrong calendar data.**

Evidence read: the full 1,531-line diff across 16 files at `edb27615`, plus `gh pr checks 160`. `checks` and `e2e` pass; `check` fails. I read the failed job log: `db::migrate::registration_guard::versions_are_dense_and_match_their_filenames` reports `0058_google_calendar_sync is out of order — expected version 57`; result `1954 passed; 1 failed`.

Additional blockers visible in the current code:

- `calendar_accounts.oauth_refresh_token` stores refresh-token plaintext in SQLite even though live calls reread the token file. This creates an unnecessary second secret copy.
- Config/token loading reads `HOME` directly instead of the amux-home resolver. The E2E sets `AMUX_HOME`, so the test can read a developer's real home/config instead of its fixture.
- Account token/fetch failures are skipped and the response still returns an ordinary `events`/`total` payload. Google pagination is not followed (`nextPageToken` is only logged, `maxResults=250`), so partial data is presented as complete with no per-account verdict, `measured`, or `n_considered` signal.
- Provider status becomes uppercase `CANCELLED`, but the UI filters lowercase `cancelled`, so cancelled events render. All-day `YYYY-MM-DD` is rewritten to `YYYYMMDD` before FullCalendar receives it.
- The title/body still promise update/delete/iCal and full sync although the current branch was re-scoped to live read plus create. That stale contract should be corrected.

The existing E2E does not exercise real Google pagination, account failure, secret isolation, cancellation, or all-day rendering, so its green result cannot detect these defects. Owner/next action: @ivomarino to renumber against current main, remove the plaintext token copy, use the amux-home resolver, make partial-account/page outcomes explicit, fix event normalization, and update the PR contract/tests. This touches startup/schema/shared calendar paths and will need a focused maintainer review afterward.

esteininger (changes requested):
Reviewed the full 16-file `gh pr diff 160` at f171e7ef117be2b81f0776a4bcf219325dcbb7e0 and `gh pr checks 160`. **Request changes; leave open.** All three checks are now green; the old branch-local 57/58 density failure is no longer a valid finding.

Current code findings:

- `gcal_sync::convert_event` returns `CANCELLED`, while the dashboard filters `e.status === 'cancelled'`. Add a conversion-to-dashboard assertion; the current unit test only pins `CONFIRMED`.
- `parse_google_datetime` strips ISO all-day dates into `YYYYMMDD`, then the dashboard passes them to FullCalendar. Its unit test pins that transformed string, without testing the calendar consumer. Preserve the consumer's date contract and prove an all-day event renders correctly.
- `calendar_init::initialize_accounts` copies refresh tokens into the plaintext `calendar_accounts.oauth_refresh_token` column even though API reads reload the connector token. Remove the redundant credential copy. Both config/token loaders also resolve directly from HOME rather than the configured amux home.
- Event requests stop at `maxResults=250`; neither event nor calendar listing follows `nextPageToken`. Individual calendar/account failures can become a successful empty/partial aggregate. Add paginated/malformed/upstream-failure fixtures and an explicit completeness/error verdict alongside the count, with durable logs for dropped work.

The current implementation is read + create without a local event mirror, periodic sync, update/delete, or a second iCal feed. Please rewrite the title/body around that actual scope. @esteininger owns the remaining product choice: a Google read layer only, or read + explicit Google event creation alongside the existing calendar. I have not approved either on his behalf.

Also, main 7321e562 now contains migrations 57–60; this branch's 57 collides with main and #163. Coordinate renumbering when integrating, without reserving a gap in either independent PR. This is shared numbering coordination, not evidence that your earlier choice was wrong.

Owner: @ivomarino for the code and regression coverage; Ethan for the product decision. No calendar data was modified during review.

esteininger (commented):
Reviewed, and the important design question here is already resolved correctly: earlier revisions built a full parallel calendar system (a local event mirror, a periodic sync job, a second iCal feed), which would have duplicated the existing /api/calendar.ics primitive (CLAUDE.md's iCal sync section, and ethos rule 5 on accumulating vs. discriminating). That got caught and re-scoped down to account storage plus live pass-through read/create against Google directly, no mirror, no second feed. The existing calendar.ics stays the one source of truth for events. Good call, and the header comments in api/gcal.rs and db/calendar.rs documenting that history are appreciated, saved me from re-deriving the same conclusion the hard way.

Two things before merge:

1. This needs a rebase, currently  against current main (the usual next-free-slot collisions in mod.rs/migrate.rs registration and app.js/sw.js version bumps, nothing structural).

2.  is a NOT NULL column that calendar_init.rs writes the plaintext refresh token into at startup, but nothing ever reads it back, the actual token lookups re-read the JSON file at ~/.amux/connectors/google/<email>.json directly (same file gmail_auth.rs already uses). That's a dead, unnecessary second plaintext copy of a live credential sitting in SQLite. Please either drop the column and the write, or actually wire it up as the source of truth, whichever you'd rather do, but not both existing at once.

CI is green, tests look reasonable for the scope. Once those two are addressed this looks mergeable to me.

esteininger (commented):
Reviewed, and the important design question here is already resolved correctly: earlier revisions built a full parallel calendar system (a local event mirror, a periodic sync job, a second iCal feed), which would have duplicated the existing /api/calendar.ics primitive (CLAUDE.md's iCal sync section, and ethos rule 5 on accumulating vs. discriminating). That got caught and re-scoped down to account storage plus live pass-through read/create against Google directly, no mirror, no second feed. The existing calendar.ics stays the one source of truth for events. Good call, and the header comments in api/gcal.rs and db/calendar.rs documenting that history are appreciated, saved me from re-deriving the same conclusion the hard way.

Two things before merge:

1. This needs a rebase, currently `mergeable: CONFLICTING` against current main (the usual next-free-slot collisions in mod.rs/migrate.rs registration and app.js/sw.js version bumps, nothing structural).

2. `calendar_accounts.oauth_refresh_token` is a NOT NULL column that calendar_init.rs writes the plaintext refresh token into at startup, but nothing ever reads it back. The actual token lookups re-read the JSON file at ~/.amux/connectors/google/<email>.json directly (same file gmail_auth.rs already uses). That's a dead, unnecessary second plaintext copy of a live credential sitting in SQLite. Please either drop the column and the write, or actually wire it up as the source of truth, not both existing at once.

CI is green and tests look reasonable for the scope. Once those two are addressed this looks mergeable.

PR #166, opened 2026-08-27
feat: add systemd user service support for Linux

ivomarino:
Comprehensive systemd user service support for Linux (Ubuntu 22.04+, Debian 11+, Fedora 36+) with auto-rebuild timer, complete documentation, and README integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,42 @@ make test       # clippy + cargo test
 
 `make run` is the command after `git pull` — it rebuilds release, installs the binary, and the launchd-managed server picks it up automatically. `make dev` is for working on migrations or features you don't want touching the live DB.
 
-**Requirements:** macOS (primary; on Linux the installer builds and installs the binaries and prints how to run the server), tmux 3.2+, and at least one of Claude Code, Codex CLI, or Gemini CLI. The Rust toolchain is installed via rustup if you don't have it (with your confirmation).
+**Requirements:** tmux 3.2+, and at least one of Claude Code, Codex CLI, or Gemini CLI. The Rust toolchain is installed via rustup if you don't have it (with your confirmation).
+
+### Linux: systemd user services
+
+On Linux with systemd (Ubuntu 22.04+, Debian 11+, Fedora 36+), `./install.sh` automatically creates and enables three user-level services:
+
+- `amux-server.service` — the main server
+- `amux-builder.service` — auto-rebuild on code changes  
+- `amux-builder.timer` — periodic rebuild check (every 60s)
+
+After `./install.sh` completes, the services are ready to start:
+
+```bash
+systemctl --user enable amux-server amux-builder.timer
+systemctl --user start amux-server
+```
+
+View logs and status:
+
+```bash
+journalctl --user -u amux-server -f      # follow logs
+systemctl --user status amux-server      # service status
+```
+
+See [docs/systemd-setup.md](docs/systemd-setup.md) for complete documentation: troubleshooting, multi-user setup, environment overrides, and migration between versions.
+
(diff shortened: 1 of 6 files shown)


esteininger:
**Review, not a merge — but this is close, and two specific fixes get it there.**

Checked first, because it is what would have blocked it outright, and it holds: **macOS is provably untouched.** The new branch is gated `[[ "$OS" == "Linux" ]] && command -v systemctl` and `exit 0`s before the existing `!= "Darwin"` block, so the launchd path is unreachable from it. The templates are clean too — every value is `$BIN_DIR` / `$PORT` / `$AMUX_HOME` / `$SCRIPT_DIR` / `$HOME` substituted at install time, no machine-specific paths, and `CARGO_TARGET_DIR=$HOME/.amux/rust-build-target` matches the one-shared-build-dir rule rather than inventing a per-service tree. CI green including the Rust `check`.

It also resolves the old comment honestly rather than around it. `install.sh` used to say "pretending to manage systemd from a bash installer is how services half-exist" — this doesn't pretend, it writes real units, runs `daemon-reload`, and `die`s on failure.

## 1. The builder unit will run 6x more often than intended

`rust-auto-build.sh` is a **one-shot**. It has no loop — macOS drives it with launchd `StartInterval=60` + `RunAtLoad`, i.e. re-invoked every 60 seconds.

The template translates that as a daemon:

```ini
Type=simple
Restart=always
RestartSec=10
```

plus `amux-builder.timer` with `OnUnitActiveSec=60s`. For a script that exits immediately, `Restart=always` + `RestartSec=10` **is** the loop, at 10 seconds — and the timer is both redundant and fighting it, because it tries to start a unit that systemd is already restarting. On a script that git-fetches and can kick a cargo build, a 6x rate increase is not cosmetic.

The one-shot-plus-timer form is:

```ini
[Service]
Type=oneshot
ExecStart=$SCRIPT_DIR/rust-auto-build.sh
# no Restart=, no [Install] — the TIMER is what gets enabled
```

and drop `[Install] WantedBy=default.target` from `amux-builder.service` so `systemctl --user enable amux-builder` can't accidentally install both alongside the timer. `amux-server.service` is correctly `Type=simple` + `Restart=always` — that one really is a daemon, and the asymmetry between the two units is the point.

## 2. `StandardErrorStream=journal` is not a systemd directive

In `amux-server.service.template`:

```ini
StandardError=journal
StandardErrorStream=journal   # <- not a thing
```

`systemd.exec(5)` has `StandardOutput=`, `StandardError=`, `StandardInput=`. systemd will log `Unknown key name 'StandardErrorStream' in section 'Service', ignoring` on every `daemon-reload`. Harmless, and it duplicates the line above it, so it just wants deleting.

## 3. Small, not blocking: `envsubst` is unchecked

The installer checks `cargo`, `tmux` and `herdr` with `command -v` before using them. `envsubst` gets used without one, and it ships in `gettext-base` (Debian/Ubuntu) / `gettext` (Fedora), which is not always present on a minimal install. The `|| die "failed to create amux-server.service"` fires, but names the template rather than the missing package. One line in the file's existing idiom fixes it:

```sh
command -v envsubst >/dev/null || die "envsubst required (apt install gettext-base / dnf install gettext)"
```

## Noted, not a request

The Linux path `exit 0`s before section 6, so the installer never waits on `/health` there. That reads correct to me — it does not START the service, it prints `systemctl --user enable/start`, so there is nothing to wait for. Worth being deliberate about rather than incidental: on macOS the installer confirms the server came up, and on Linux it cannot make that promise. Saying so in the final output would keep the two honest with each other.

Fix 1 and 2 and I'll merge this. The gap it closes is real — the installer currently tells Linux users "no service manager configured" — and the shape is right.

PR #168, opened 2026-08-28
fix(build): cap cargo's parallelism to jobs=1 — mitigates repeated session crashes

ivomarino:
## What

`install.sh`'s generated `.cargo/config.toml` now sets `[build] jobs = 1`, serializing every cargo invocation in this checkout instead of the default `nproc` (4).

## Why

Over one working session, running `cargo test -p amux-server` (compiles + links the whole ~1466-test binary) correlated with 3-4 session-ending kills on the dev box — one killed process was the session's own foreground shell, not just the build. `jobs=2` was tried first and was **not** enough; three more crashes followed it in the same session. `jobs=1` is the one value this investigation actually confirmed held.

**Honest caveat:** the exact resource ceiling triggering this is *not* pinned down. An earlier pass of the investigation claimed a specific 2.75GiB cgroup memory cap — that was a self-caught analysis error (`memory.current`, current usage, misread as `memory.max`, the limit) and has been retracted. Full trail on board card FRONT-2 in the fleet's own tracker. This PR is a mitigation based on observed correlation, not a confirmed root-cause fix, and the comments in both `install.sh` and the generated config say so plainly rather than overclaiming.

## Verified

- `git diff` reviewed by eye (single-file, additive-only change to a generated-config template)
- Confirmed via a fresh `git clone` that the committed content matches intent (not just the working tree)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -130,8 +130,27 @@ else
 #
 # Lowest precedence of cargo's three: CARGO_TARGET_DIR and --target-dir still
 # override it, which is what keeps the e2e worktree build and CI unaffected.
+#
+# jobs cap (FRONT-2, 2026-08-28): cargo's default parallelism (= nproc, so 4
+# rustc/linker jobs at once for amux-server, a large ~1466-test crate)
+# correlates with repeated session crashes on this box — every tmux session
+# in a shared checkout's fleet lives in the same container, so a kill under
+# memory pressure is not necessarily the build's own process; it can reap an
+# unrelated session's Claude Code process as collateral, which is what made
+# this read as "the worker randomly crashed" for months instead of "this
+# build ran out of room". EXACT MECHANISM NOT FULLY PINNED: an earlier pass
+# of this investigation claimed a specific 2.75GiB cgroup cap, which was a
+# misread (memory.current, current usage, mistaken for memory.max, the
+# limit) and has been retracted (see board card FRONT-2) — this container's
+# own cgroup shows no configured max. A host-level limit outside what is
+# visible from inside the container remains plausible but unverified. What
+# IS verified: jobs=2 was tried first and was NOT enough (three more crashes
+# followed in the same session). Serial (jobs=1) is the one value this
+# investigation has actually confirmed holds; revisit upward only with real
+# headroom evidence, not a guess.
 [build]
 target-dir = "$SHARED_TARGET_DIR"
+jobs = 1
 EOF
   say "cargo config: builds in this checkout target $SHARED_TARGET_DIR"
 fi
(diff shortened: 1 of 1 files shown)


esteininger:
**Review, not a merge.** The investigation behind this is better than most bug reports I read here — you tried `jobs=2` first, it wasn't enough, you say so; and you caught and retracted your own `memory.current` vs `memory.max` error rather than shipping a confident wrong cause. The "mitigation based on observed correlation, not a confirmed root-cause fix" framing is exactly right and I'd rather have that than a tidier story.

The problem is where the value lands, not whether it helped you.

## `jobs` is machine-specific, and this file is the machine-specific one

`install.sh` generates `.cargo/config.toml`, and that file's own header explains why it is gitignored:

```
# GENERATED by install.sh (AMUX-3667). Gitignored on purpose — the path is
# absolute and machine-specific. Re-run ./install.sh to regenerate.
```

`target-dir` is there because it cannot be committed. `jobs = 1` is the same kind of value for the same reason: the right number depends on the box's memory and core count. Your evidence is from a memory-constrained environment (you're reading cgroup `memory.*`, so Linux, containerised). On the machine this fleet actually runs on, `sysctl hw.ncpu` reports **28**.

So this hardcodes a fact about one constrained box into every install. That is a shape this codebase has a name for and has fixed twice recently — the ollama model literal (`DESKT-6`) and, in PR #163, a hardcoded `~/.config/sops/age/keys.txt`.

## The blast radius is bigger than a manual `cargo test`

The pre-commit hook runs `cargo check --workspace --all-targets` on **every commit**, and ~50 lanes share this checkout. Serialising that takes the gate from 28-way to 1-way for all of them, on a box that has no memory pressure. That is a large, permanent, fleet-wide cost imposed to fix a condition that does not exist there.

## What I'd merge instead

Same fix, made conditional, in the shape this repo already uses for exactly this class of number (`AMUX_HELPER_MODEL`, `AMUX_OLLAMA_DEFAULT_MODEL`, `AMUX_PICKUP_EXCERPT_CHARS` — all env-overridable because a policy constant compiled in becomes the ceiling, deviation D4):

```sh
# in install.sh, when generating .cargo/config.toml
jobs="${AMUX_CARGO_JOBS:-}"
[ -n "$jobs" ] && printf 'jobs = %s\n' "$jobs" >> "$cfg"
```

Then your box runs `AMUX_CARGO_JOBS=1 ./install.sh` and gets the serialisation it needs, and nobody else pays for it. Keep the comment you wrote — it is good, and it belongs next to the knob.

If you'd rather it default to 1 on constrained hosts specifically, detecting a cgroup memory limit and setting it only there would also be mergeable; that keeps the value derived from the machine rather than from a checkout.

## One thing worth chasing separately

"One killed process was the session's own foreground shell, not just the build" is a bigger deal than the cargo setting, and `jobs=1` would hide it rather than fix it. If an OOM killer is reaping a shell on that box, the cargo parallelism is the trigger and not the fault. Worth its own issue with the `dmesg`/`journalctl` evidence — I'd read that one.

PR #170, opened 2026-08-29
feat(telegram): bot connector — long-poll inbound, REST outbound, chat/session linking

ivomarino:
## Summary

Adds a Telegram bot connector so an amux session can be reached from, and reply to, Telegram — long-polling for inbound (no public IP/webhook needed, works from behind NAT), a REST endpoint for outbound.

- **migration**: `telegram_mappings` (chat_id <-> session) + `telegram_poll_state` (getUpdates offset checkpoint, durable across restarts)
- **`db::telegram`**: plain CRUD, no state-event journal — this is operator config, not fleet-wide fact traffic
- **`runtime_jobs::telegram_poll`**: long-poll loop, idles cleanly when `TELEGRAM_BOT_TOKEN` is unset. `/link <session>` self-registers a chat (session existence validated first); other text from a linked chat routes into the session stamped `[from Telegram @user]:`; an unlinked chat gets a one-line nudge back, never silently dropped
- **`api/telegram`**: `GET /status`, `GET`/`POST /mappings`, `DELETE /mappings/{chat_id}`, `POST /send`
- **connector registry**: Telegram added as an `ApiKey` provider (bot token only, no OAuth, no webhook URL to configure)
- **skill docs**: `skills/amux.md` (mirrored to `.claude/commands/amux.md`) documents the full command set

## Fix included

A second commit fixes a bug found testing this live: several write paths (`upsert`, `remove`, `touch_last_message`, the offset checkpoint) were going through `state.store.read()` instead of `state.store.write_async()`. This codebase's read-pool connections have `query_only=ON` set permanently by design (`db/mod.rs`'s single-writer-thread architecture) — every one of those writes failed 100% of the time with "attempt to write a readonly database", independent of restarts. All five call sites now go through the single writer thread, matching the pattern every other db module uses.

## Testing done

Verified live end-to-end against a real bot: `/link <session>` from Telegram succeeds and persists, inbound text routes into the linked session with correct attribution, an unlinked chat gets the nudge, `POST /api/telegram/send` delivers outbound, and the manual `POST`/`DELETE /api/telegram/mappings` pair (used by the skill docs) both succeed with correct status codes (200/204/404).


esteininger:
Core feature looks solid and CI is green, but there are dev artifacts in the diff that need to be removed before merge:

- `.claude/settings.json`
- `.claude/telegram-relay.py`
- `CLAUDE.local.md`
- `Dockerfile.remote-build`
- `amux-server.new`

Strip those from the branch and force-push, then this is ready to merge.

ivomarino:
All 5 addressed:

- `CLAUDE.local.md`, `amux-server.new` — removed earlier today (were also scrubbed from history, not just HEAD — they'd been pushed to this fork before I caught it)
- `.claude/telegram-relay.py` — deleted. Fully superseded by the server-side `telegram_relay` runtime job added this PR (works for every session with zero per-hook config, unlike the hook it replaced — confirmed working end to end against real Telegram traffic multiple times today)
- `.claude/settings.json` — reverted to byte-identical with `main`. The only diff this branch ever introduced was the one Stop-hook entry invoking the now-deleted `telegram-relay.py`
- `Dockerfile.remote-build` — deleted. Personal single-machine dev tooling, never should have shipped

Pushed as a normal commit rather than a force-push — this repo squash-merges (confirmed on #166/#168/#162/#161), so the intermediate branch history collapses at merge time regardless of whether it's rewritten now. Happy to force-push instead if you'd rather the raw commit list be clean too, just say so.

PR #177, opened 2026-09-02
fix(autofix): AMUX-97 -- POST /api/browser/start is LONG_BY_DESIGN, budget 30s

ivomarino:
Card AMUX-97 filed on a single 10.3s `POST /api/browser/start` call (n=1, no baseline in 7 days), arriving while this box's 1-minute load was 1.96x on 4 cores (concurrent compile/test). Per the card's own caution — "if the duration is by design, the fix is a LONG_BY_DESIGN entry derived from the route's own bounds, not a threshold nudge" — this is that fix.

**Root cause (instrument, not the route):** `integrations/browser.rs::start` polls Chrome's CDP HTTP port for up to 30s before giving up, deliberately widened past the old 12s because "a real logged-in profile on a busy machine (a 50-lane fleet) routinely takes longer than the old 12s to open the debugging port." 10.3s is comfortably inside that documented 30s launch budget — the detector's flat 10s floor is below this route's own normal, the exact threshold-below-baseline shape the card warns about.

Same pattern as the existing `/api/browser/navigate` entry (AMUX-3835) two lines above it in the table: derive the budget from the endpoint's own timeout rather than guessing, and add a dedicated test that pins the *lookup*, not just the table entry, so it only stays green while the detector can actually reach it. Past 30s the CDP wait itself failed to bound the call — that's still a real hang and still files.

### DIFF
--- a/crates/amux-server/src/runtime_jobs/autofix.rs
+++ b/crates/amux-server/src/runtime_jobs/autofix.rs
@@ -188,6 +188,24 @@ const LONG_BY_DESIGN: &[(&str, f64)] = &[
     // design bound (40s), which is the threshold-below-baseline defect the
     // card's own closing line names.
     ("/api/browser/navigate", 40_000.0),
+    // POST /api/browser/start (AMUX-97). The wait is a real Chrome launch: the
+    // handler polls CDP's HTTP port for up to 30s before giving up
+    // (integrations/browser.rs `start`, `deadline = now + 30s`), and that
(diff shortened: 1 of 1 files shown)


ivomarino:
Woodpecker CI verified green (clone/check/clippy/test all passing) on `1c78898b`. This is ready for review/merge.

ivomarino:
@esteininger this one's ready for your review — small, self-contained, CI green. Summary is in the PR description and the comment above.

PR #178, opened 2026-09-02
fix(browser): AMUX-98 -- a malformed selector is 400, not 502

ivomarino:
Card AMUX-98 filed on `POST /api/browser/action` → 502 with body `eval: SyntaxError: Failed to execute 'querySelector' on 'Document': 'text=Create credentials' is not a valid selector.` The caller used Playwright-style selector syntax (`text=...`) against amux's native CDP driver, which speaks plain CSS via `document.querySelector`. Per the card's own text: "A 5xx is amux failing, not amux declining — if this is really a refusal, the fix is the STATUS CODE, not the caller."

**Root cause:** `CdpClient::eval()` folded two very different failure classes into one untyped `anyhow::Error` — a page-side JS exception (the script, or a caller-supplied selector interpolated into one, threw) and a real CDP transport/protocol failure. `click_selector`, `set_input_files`, and the "wait" selector probe all interpolate a caller-supplied selector into an `eval()` call and propagated whichever came back as a bare `Err`, so a malformed selector 502'd exactly like a wedged browser — the same row-folding defect AMUX-3672 already fixed once (`CdpTimeout` vs a protocol error → 504 vs 502), recurring here as a third unhandled failure class.

**Fix:** same technique as AMUX-3672 — carry the page exception as a TYPE (`PageException`, `integrations/browser.rs`), not a string. `cdp_status()` now downcasts three ways: `CdpTimeout` → 504, `PageException` → 400, everything else → 502. Applied uniformly across every `Err(e)` arm in `api/browser.rs`'s `action` dispatch (click, type, input, key, scroll, files, back, wait, viewport) rather than just the one arm that filed the card, since `eval()` is their shared mechanism and click/files/wait all interpolate a caller selector into it the same way.

Message text is byte-for-byte unchanged (`"eval: {desc}"`), so nothing reading the error body sees a difference — only the type is now inspectable, same non-breaking shape as the `CdpTimeout` precedent.

**Tests:** `a_page_exception_is_400_not_502` pins the 3-way `cdp_status()` dispatch plus two CONTROL cases (a real protocol error stays 502, a `PageException` wrapped in `.context()` is still recognised) — matches the existing `a_cdp_timeout_is_504_and_a_protocol_error_is_502` test's own depth.

### DIFF
--- a/crates/amux-server/src/api/browser.rs
+++ b/crates/amux-server/src/api/browser.rs
@@ -616,21 +616,30 @@ fn driver_err(e: chrome::DriverError) -> Response {
     }
 }
 
-/// 504 for a CDP call that never answered, 502 for one that failed (AMUX-3672).
+/// 504 for a CDP call that never answered, 400 for a page-side exception
+/// (AMUX-98), 502 for everything else — a real transport/protocol failure
+/// (AMUX-3672).
 ///
-/// Both were 502, and `/api/logs/analyze` groups by (status, method, target) —
-/// so "the browser is WEDGED" and "the browser REJECTED our call" landed in the
-/// same row and only the error body separated them. A wedged browser is the one
-/// that means something else is wrong (a contended profile, a hung tab); a
-/// protocol error usually means the caller asked for something impossible.
-/// Making them different status codes makes them different groups in every log
-/// view, for free.
+/// All three used to be 502, and `/api/logs/analyze` groups by (status,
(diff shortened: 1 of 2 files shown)


ivomarino:
Woodpecker CI verified green (clone/check/clippy/test all passing) on `62cc96d8`. This is ready for review/merge.

ivomarino:
@esteininger this one's ready for your review too — small, self-contained, CI green. Summary is in the PR description and the comment above.

PR #179, opened 2026-09-02
docs(frustrations): log AMUX-99 -- runtime hook drift, install.sh has no supervision

ivomarino:
Found live while sweeping `GET /api/health/invariants` for other reasons: `hooks.report_hook_matches_committed` and `hooks.shared_guard_matches_committed` were both failing.

**Root cause:** `~/.amux/hooks/git-shared-guard.py` and `~/.amux/hook-report.sh` were both installed 2026-08-30 and never reinstalled since, while their source (`scripts/git-hooks/git-shared-guard.py`, `scripts/hooks/hook-report.sh`) kept getting real commits — most notably `e782b68a` (AMUX-3932), a genuine guard-**bypass** fix ("command substitution inside a quoted argument bypassed the shared-checkout guard"). That fix passed every CI gate and sat in git history, never live on the box, because nothing re-runs `install.sh`'s hook-install step automatically.

This is a **recurrence** — AMUX-28/AMUX-29 already covered this exact invariant pair and are marked `done` with no evidence recorded on either. The underlying gap (`install.sh` only runs manually, unlike the Rust binary auto-builder / `amux-builder.timer`) was never actually closed.

**Fixed live** (on the box, not in this diff): manually re-ran `install.sh`'s own `install_hook_from_head` sequence for both files. Confirmed via the invariants endpoint: failures dropped from 6 to 4, both `hooks.*` entries cleared.

**This PR** is just the `frustrations.md` log entry plus a new card (AMUX-99) for the durable fix, which is a real design choice and not made here: either a systemd timer polling `install.sh`'s hook block the way `amux-builder.timer` polls the Rust build, or the invariant self-healing (it already computes the right bytes via `HEAD:<rel>`).

### DIFF
--- a/crates/amux-server/src/api/browser.rs
+++ b/crates/amux-server/src/api/browser.rs
@@ -616,21 +616,30 @@ fn driver_err(e: chrome::DriverError) -> Response {
     }
 }
 
-/// 504 for a CDP call that never answered, 502 for one that failed (AMUX-3672).
+/// 504 for a CDP call that never answered, 400 for a page-side exception
+/// (AMUX-98), 502 for everything else — a real transport/protocol failure
+/// (AMUX-3672).
 ///
-/// Both were 502, and `/api/logs/analyze` groups by (status, method, target) —
-/// so "the browser is WEDGED" and "the browser REJECTED our call" landed in the
-/// same row and only the error body separated them. A wedged browser is the one
-/// that means something else is wrong (a contended profile, a hung tab); a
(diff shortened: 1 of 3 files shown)


ivomarino:
@esteininger Woodpecker CI verified green (clone/check/clippy/test all passing) on `0cbd8b1d`. This is doc-only (frustrations.md entry + a new card for the durable-fix design decision) — ready whenever.

PR #180, opened 2026-09-02
feat(telegram): group-chat support -- read/write via @lane mention, gated linking

ivomarino:
Per your request ("have the bot write and read to groups he's in"). Two design questions were confirmed with you before building this: **mention-only trigger in groups** (not every message), and **a group can only link to a session that already has a private link**.

**What I found first:** the existing `/link` + message-routing flow never checked chat type at all — Telegram hands a group a `chat_id` exactly like a private chat, so a group could *already* be `/link`'d today, with every message from every member forwarded to the linked session verbatim. No noise control, no eligibility gate.

**What this PR adds:**
- `migrations/0054_telegram_chat_type.sql` — `telegram_mappings.chat_type` (`private`/`group`/`supergroup`), backfilled `private` for every existing row (correct by construction: nothing before this could distinguish a group, so nothing could have deliberately linked one).
- `has_private_link(session)` — the eligibility gate. `/link` in a group now refuses unless the target session already has a private link, with a reply telling the sender to link privately first.
- Once linked, a group message is delivered **only** if it starts with a valid `@lane` mention — reusing the exact `@session_name` convention already used for multi-lane routing in private chats. Everything else in the group (including an unlinked group's messages) is silently ignored — no reply, no noise. A typo'd mention in a group also gets no fallback delivery (unlike private chats), since guessing a default would forward the message to a session the sender may not have meant to address.
- The same eligibility gate is enforced on the admin API path (`POST /api/telegram/mappings`) too, not just the Telegram-side `/link` command — a gate with a bypass on one of its two entry points isn't really the gate.
- `telegram_relay.rs` (outbound) needed **zero** changes — it already watches whatever `routed_session()` resolves to and sends to `mapping.chat_id` regardless of chat type, so a reply to a group-routed message already lands back in the group correctly.

**Tests:** 5 new unit tests on the DB layer (in-memory SQLite, matches this codebase's existing pattern) covering the new column, the eligibility gate's true/false/group-doesn't-count cases, re-link behavior, and a session legitimately holding both a private and a group link at once.

Verified via Woodpecker CI (clone/check/clippy/test, all green).

### DIFF
--- a/crates/amux-server/migrations/0054_telegram_chat_type.sql
+++ b/crates/amux-server/migrations/0054_telegram_chat_type.sql
@@ -1,0 +1,21 @@
+-- Group-chat support (Ethan, 2026-09-02: "have the bot write and read to
+-- groups he's in"). `/link` and message routing never checked chat type --
+-- Telegram hands out a chat_id identically for a private chat and a group,
+-- and the existing code trusted whichever chat_id sent the update. That
+-- meant a group could already be `/link`'d today, with every message from
+-- every member in it forwarded to the linked session -- unusable noise, and
+-- no gate on who could point a group at somebody's session.
+--
+-- `chat_type` records what Telegram told us at link time ('private',
+-- 'group', 'supergroup') so the app layer can: (a) require mention-only
+-- delivery in a group (skip anything that isn't an @lane mention, instead
+-- of relaying every message), and (b) require the SESSION already have an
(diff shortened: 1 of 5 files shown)


ivomarino:
@esteininger this is the group-chat feature from earlier — both design questions (mention-only trigger, private-link-required-first) confirmed with you before building. CI green, ready whenever.

PR #183, opened 2026-09-03
docs(frustrations): archive the AMUX-70 cargo-OOM entry (validated by amux)

ivomarino:
Ran `scripts/frustrations-archive.py` earlier this session but never committed the result — caught it doing a routine cleanliness sweep, still sitting uncommitted in the working tree.

The entry ("A local `cargo clippy` OOM-kill doesn't just kill the build — it kills the WHOLE interactive session") is genuinely resolved: card AMUX-70 is verified with concrete evidence (`scripts/safe-cargo.sh check` runs in its own sibling systemd scope, a real commit went through the newly-fixed pre-commit hook with no hang), and this session independently reconfirmed it live today — every local cargo invocation (multiple `git commit`s, each triggering the pre-commit hook's check+clippy gate) went through the wrapper and never took the interactive pane down.

Moved via the sanctioned tool, which stamps `VALIDATED: amux` and carries the SYMPTOM+COST onto AMUX-70 itself so a recurrence is recognisable from the card alone.

Verified via Woodpecker CI.

### DIFF
--- a/frustrations-archive.md
+++ b/frustrations-archive.md
@@ -3766,3 +3766,70 @@ FIXED, verified 2026-09-02 by running the exact test this entry describes.
   transition will accept".
   So the sanctioned instruction no longer points at a source that will refuse you.
   Measured on a throwaway card and deleted after.
+
+## A local `cargo clippy` OOM-kill doesn't just kill the build — it kills the WHOLE interactive session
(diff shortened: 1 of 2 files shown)


ivomarino:
@esteininger small doc-only cleanup — CI green, ready whenever.

PR #186, opened 2026-09-03
fix(autofix): AMUX-112 -- GET /api/gmail/accounts is LONG_BY_DESIGN, budget 30s

ivomarino:
Third filing on the same residual (AMUX-84/90 before it). The card's own occurrence (20015ms) is comfortably inside the worst case already measured and documented in the code itself: `probe_health`'s own comment says "24.3s across 3 connected accounts" — a real production measurement, not a guess made here.

**Root cause of the recurring filing:** this route shares `/api/gmail/inbox`'s exact shape (a bounded-concurrency fix from AMUX-84, the same `ReqwestTransport` with its 30s per-call timeout in `integrations/email.rs`) but never got the matching `LONG_BY_DESIGN` entry `gmail/inbox` already has — so the detector's flat 10s floor kept firing on every legitimate slow-account health check.

**Fix:** added `("/api/gmail/accounts", 30_000.0)`, same reasoning `gmail/inbox`'s own entry already uses (matches the transport's own timeout directly, not a padded guess), plus a test that pins the lookup and asserts the entry still files past 30s (a real transport hang stays a real hang).

### DIFF
--- a/crates/amux-server/src/runtime_jobs/autofix.rs
+++ b/crates/amux-server/src/runtime_jobs/autofix.rs
@@ -233,6 +233,24 @@ const LONG_BY_DESIGN: &[(&str, f64)] = &[
     // exactly what a design-bounded quota wait looks like from outside.
     // 30s matches /api/email/inbox's own reasoning (2x worst measured).
     ("/api/gmail/inbox", 30_000.0),
+    // GET /api/gmail/accounts (AMUX-112, third filing on the same residual —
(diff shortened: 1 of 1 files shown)


ivomarino:
@esteininger Woodpecker CI verified green (clone/check/clippy/test all passing, including the new pinned-lookup test). Ready whenever.

ivomarino:

PR #187, opened 2026-09-03
fix(backup): restore scripts/amux-credential-backup.sh -- silently lost, service failing every run

ivomarino:
Found while answering an infra peer's questions about migrating the shared box: `amux-credential-backup.service` (AMUX-76 — backs up single-copy credential files, specifically `~/.amux/gmail-oauth-client.json`, whose earlier unexplained disappearance is exactly what this service exists to guard against recurring) has been **failing on every single 6h run** since shortly after it was first installed:

```
systemd[...]: Unable to locate executable '/home/syseng/src/amux/scripts/amux-credential-backup.sh': No such file or directory
systemd[...]: Main process exited, code=exited, status=203/EXEC
```

The script was genuinely committed once (`a41dea63`, 2026-09-01) and its systemd unit + timer are both still correctly installed and enabled — but the script itself is absent from current `main` entirely, with **no delete commit** in its history. Most likely dropped silently during one of this session's many large multi-way merges rather than deliberately removed (the "clean merge, zero conflict markers, still wrong" class of defect `CLAUDE.local.md` already documents from a separate incident).

**Real exposure**: only ONE backup ever succeeded (`2026-09-01T11:04:23Z`, the very first run); every run since has 203/EXEC'd silently into the journal — nothing surfaces a `Type=oneshot` failure anywhere a session would see it by default. ~2 days with zero backup protection on a file whose own commit message says plainly "no other recovery path." The protected file itself is still present on disk (confirmed), so nothing was actually lost — but the safety net covering it was gone the whole time.

**Fix**: restored the script byte-for-byte from `a41dea63` (no drift to reconcile). Ran it live post-restore: `backed up 1/1 file(s)` — confirmed a fresh timestamped backup now exists, closing the exposure window immediately rather than waiting for the next scheduled tick.

### DIFF
--- a/scripts/amux-credential-backup.sh
+++ b/scripts/amux-credential-backup.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+# Periodic backup of single-copy credential-shaped files under ~/.amux —
+# files this box has NO other recovery path for if they vanish (AMUX-76:
+# ~/.amux/gmail-oauth-client.json disappeared from disk with root cause
+# never determined; recovery only worked because its client_id/secret
+# happened to ALSO live in server.env under a different key — a lucky
+# coincidence this script exists so the next file doesn't need).
+#
+# Deliberately narrow: only files known to be (a) single-copy — nothing
+# else on this box holds their content, and (b) not already covered by
+# git or another backup path. Add to this list as new single-copy
+# credential files are introduced; do NOT widen it to "everything in
+# ~/.amux" — most of that is either regenerable, DB-backed (already has
+# its own durability story), or explicitly gitignored on purpose.
+#
+# No secret VALUES appear in this script or its own output — only paths.
+set -euo pipefail
+
+AMUX_HOME="${AMUX_HOME:-$HOME/.amux}"
+BACKUP_DIR="$AMUX_HOME/backups/credentials"
+KEEP=10 # prune to the last N backups per file, so this can't grow unbounded
+
+FILES=(
+  "$AMUX_HOME/gmail-oauth-client.json"
+)
+
+mkdir -p "$BACKUP_DIR"
+chmod 700 "$BACKUP_DIR"
+
+ts="$(date -u +%Y%m%dT%H%M%SZ)"
+backed_up=0
+for f in "${FILES[@]}"; do
+  name="$(basename "$f")"
+  if [ ! -f "$f" ]; then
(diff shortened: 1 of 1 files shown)


ivomarino:
@esteininger Woodpecker CI verified green. This one's worth a look sooner rather than later — it's a silently-broken safety net for a single-copy credential file, not just a latency/instrument fix.

ivomarino:
@esteininger this is mergeable; `check`/`checks` pass. `e2e` shows one flaky failure unrelated to this diff (a pure bash script restore) — retriggering it earlier today brought it down from 4 failures to 1, in a different test each time, consistent with shared-checkout CI flakiness rather than a real regression from this change. Flagging rather than re-retriggering further, since you'll want to judge that yourself.

ivomarino:
@esteininger following up with higher urgency: the fix in this PR keeps getting undone on the shared checkout (it only lives on this branch, not `main`), and today's re-check found it recurred a SECOND time — plus something more severe underneath: the actual credential file this backup protects (`~/.amux/gmail-oauth-client.json`, AMUX-76's own "no other recovery path" file) had also gone missing from disk. Recovered from a backup with no data loss, but its own root cause is undetermined (volatile journal, no log evidence survives). Filed as AMUX-124 for the credential file itself; this PR is the fix for the backup service that would have caught it faster.

Given it's now a real (if narrowly avoided) credential-exposure pattern rather than a one-off, flagging this as worth merging sooner rather than later.

esteininger:
Daily review verdict: **leave open; restoration is necessary, but the restored script still reports success when protection is absent.**

Evidence read: the complete 52-line diff at `976fb588`; current checks are `check` pass, `checks` pass, and `e2e` pending.

Root blocker: when a configured credential file is missing, the loop prints `MISSING` and `continue`s, then the script reaches its final summary and exits 0. That makes the systemd oneshot green precisely when it backed up `0/1` protected files. The PR describes a prior silent failure, but this restores a second silent-success path. `set -e` does not change the explicit `continue`.

The diff also has no automated negative check proving that a missing protected file makes the service fail, nor a positive/pruning check. The live `backed up 1/1` run proves only the happy path.

Owner/next action: @ivomarino to accumulate missing/copy failures, emit a final structured count/verdict, exit nonzero if any configured source was not protected, and add a shell test for missing, successful, and retention paths. Then wait for `e2e` to complete.

PR #188, opened 2026-09-03
fix(autofix): AMUX-114 -- GET /api/gmail/thread/{id} is LONG_BY_DESIGN, budget 30s

ivomarino:
Filed as a 3-endpoint rollup (`gmail/accounts`, `gmail/inbox`, `gmail/thread/{id}`) because all three exceeded the outlier threshold in the same 6h window — **not** one server-wide incident (`arrived_into_process_life` spanned 828.8s to 13671.0s into the process's life, ruling out a restart; host load was median 0.29x). Re-checked directly: the occurrences are spread across the whole window at genuinely different timestamps, each independently matching the same "isolated call, fast siblings" shape already established for the other two members of this family.

`gmail/accounts` already got its own `LONG_BY_DESIGN` entry today (AMUX-112, #186, not yet merged). This fixes the third and last unbudgeted member: `gmail/thread/{id}`'s own `get_thread` (`integrations/email.rs`) fetches through the identical `self.api(...)` quota-backed call the whole gmail family shares, so it gets the same 30s budget for the same reason.

**Also found and documented** (not fixed here, to keep this scoped): after the thread GET, `get_thread` clears UNREAD on every unread message **sequentially**, not with the bounded concurrency AMUX-66/84 already gave the sibling N+1 shapes on `/gmail/inbox` and `/gmail/accounts`. Both real occurrences investigated were ~30001ms/~20020ms — consistent with ONE call hitting the transport timeout, not several stacking toward 60s+ — so the flat 30s budget matches what's actually been observed. Named explicitly in the `autofix.rs` comment so it isn't lost: a thread with many unread messages could still legitimately exceed 30s without any single call hanging, which would show up as a false-positive design-duration filing here rather than a real wedge — the tell that this loop needs the same fix.

**Test**: `gmail_thread_is_judged_against_its_own_transport_timeout`, which also pins that a **real** thread id in the logged path (not the literal `{id}` template) resolves through route-table normalization to the `LONG_BY_DESIGN` entry — verified against `request_log.rs`'s own `ROUTE_TABLE` entry for this wildcard-parameter route before writing the key, not assumed.

Verified via Woodpecker CI.


ivomarino:
@esteininger Woodpecker CI verified green (clone/check/clippy/test all passing, including the route-normalization test). Ready whenever.

ivomarino:

esteininger:
Daily review verdict: **leave open; do not merge from this sweep.**

Evidence read: this PR moved during review, so I re-ran `gh pr diff 188` and read the current 89-line diff at `32671119`, not the earlier conflicting head. It is now mergeable, but the current `gh pr checks 188` snapshot has `check`, `checks`, and `e2e` all pending.

The test would catch route-normalization drift and would catch suppression extending beyond 30 seconds: it proves a real thread id at 20.02s is suppressed and 45s still files. It does **not** catch the known root cause called out in the diff: sequential unread-clearing requests can legitimately accumulate past 30s and be filed as a false wedge. This patch changes the detector budget while leaving that route-dependent false-positive mechanism in place.

It also changes `runtime_jobs/autofix.rs`, a fleet-wide shared path that should get a focused review rather than a tail-of-sweep merge. Owner/next action: @ivomarino to either bound the unread-clearing work or provide measured evidence for a budget that covers the full designed path; maintainer to review after the pending checks settle.

PR #189, opened 2026-09-04
docs(connectors): flag Slack user-identity + customer allow-list gap

ivomarino:
AMUX-3104 as currently designed would build a bot-identity Slack connector, but the real first use case (Ivo, customers lane) needs posting **as the authorizing human user's own identity** into a **customer's** own workspace (IWF's `iwf-ops.slack.com`), which the design doc doesn't cover yet. Filing this before any code lands so it isn't built bot-only by default.

Three gaps, each verified directly against the checked-out code (not assumed):

1. **Bot scopes vs. user scopes.** The registry row (`connectors.rs`) requests only bot scopes (`scope=channels:read chat:write channels:history`) — confirmed by the code's own "Slack bot tokens do not expire" comment. Posting as the user needs Slack's separate `user_scope=` OAuth param, and the callback needs to capture `authed_user.access_token` (`xoxp-...`), stored distinctly from any bot token — no such second-token shape exists in the data model today.
2. **Write-safety is deny-list-only.** The existing `deny_channels`/`deny_recipients` policy has no allow-list counterpart, so IWF's workspace needs a scoped, explicit exception (tied to the IWF CRM/customer record) rather than either silently blocked by the deny-customer-channel default, or a blanket loosening of it for Slack generally.
3. **No send/read verbs exist yet at all**, independent of the above — grepped the whole `amux-server` crate for `chat.postMessage`, zero hits. Today's Slack connector is OAuth handshake plus the `auth.test` canary only.

Docs-only, no code changes. Board card CUSTO-2 tracks this on the customers-lane side.

### DIFF
--- a/docs/design/connectors.md
+++ b/docs/design/connectors.md
@@ -13,6 +13,51 @@ Status: proposed (AMUX-3101). Owner: `amux`. Captured from Ethan's 2026-08-14 vi
 > `~/.amux/gmail-tokens/` so the email subsystem is repaired by the same approval.
 > Token rot auto-files one board card per account (autofix `connector-auth` detector).
 
+> **2026-09-04 update — Slack connector requirement gap, found before any code
+> landed:** the real driving use case for Slack (Ivo, `customers` lane) is
+> **posting as the authorizing human user's own Slack identity**, inside a
+> **customer's** own workspace (IWF's `iwf-ops.slack.com`, for ongoing managed-
+> services communication) — not a bot/app identity in an internal workspace.
+> This changes AMUX-3104's shape in two ways the rest of this doc doesn't yet
+> cover, verified directly against the checked-out code:
+> - **Bot scopes vs. user scopes.** The current `slack` registry row
+>   (`connectors.rs`) requests only bot scopes (`channels:read chat:write
+>   channels:history`) via `scope=` — confirmed by the code's own comment,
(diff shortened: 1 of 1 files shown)


ivomarino:
@esteininger docs-only, from the customers lane (Ivo's real IWF Slack use case) — flagging AMUX-3104's design gap before any code lands. Woodpecker verified green. Ready whenever.

ivomarino:

esteininger:
Daily review verdict: **leave open; route the product/security choice to Ethan.**

Evidence read: the full 103-line, docs-only diff at `4f47f306`; `gh pr checks 189` reports the sole `checks` job passing.

The document accurately identifies three real gaps: Slack currently requests bot scopes, the policy is deny-list-only, and there is no `chat.postMessage` implementation. The blocker is that the proposed resolution also decides two things the reviewer should not decide: posting as the authorizing human rather than a bot, and creating a customer-workspace allow-list exception. It also bakes a named customer's workspace into this public repository.

Options for Ethan:

1. Bot identity and internal workspaces only for the first connector.
2. Human user tokens plus a narrowly scoped customer-workspace allow-list.
3. Split the generic bot-vs-user and allow-list architecture into public docs, while keeping the named customer requirement in the customer lane/private record.

My recommendation is option 3, but that is not a merge decision I am making on Ethan's behalf. Owner: Ethan for the identity/customer-data policy; @ivomarino for the resulting doc revision.

PR #191, opened 2026-09-04
fix(e2e): worker-action-parity.spec.ts can't reach the app's own let bindings (AMUX-122)

ivomarino:
CI red on `main` since `40a108e3` (ATE-44) — `worker-action-parity.spec.ts`
fails 2/2 tests on all 3 browser projects, reproducibly, since it landed.

## Root cause (test bug, not a dashboard regression)

1. `expect(state.peekOnly).toEqual(['File browser', 'Focus mode'])` expected
   bare labels, but every menu item in this codebase renders an icon `<span>`
   before its label by design — `.textContent` legitimately includes it.
   `📂File browser`/`▴Focus mode` is what the app is *supposed* to render.

2. `boot()`'s seeding of `sessions`/`peekSession`/`peekSessionDir` (top-level
   `let` in the classic app bundle) used a nested `eval()` inside a
   `page.evaluate(fn, arg)` callback. That callback runs via Playwright's
   `Runtime.callFunctionOn` in its own isolated scope, which cannot see or
   reassign another script's `let` bindings — confirmed via the CI run's own
   `error-context.md` snapshot, which shows the exact "This worker has no
   directory to browse" toast that only fires when `_browseWorkerFiles`
   computes an empty root (i.e. the seeding never took effect). That's why
   `#path=/tmp/` never appeared in the URL.

   Fixed by using `page.evaluate`'s **string** form instead — a top-level
   `Runtime.evaluate`, which (like the DevTools console) *does* interact with
   previously-declared top-level `let`/`const`. This is already the
   established, working pattern in `peek-default-tab.spec.ts`
   (`sessions = [...]; openPeek(...)`), just not what this newer file copied.
   Same fix applied to the read-back in `enterFiles()`, which had the
   identical bug on the reading side (`_exploreSession`/`_filesPath`/
   `activeView`).

No `app.js` change — the ATE-44 feature itself works as designed; only the
test that was meant to verify it was wrong.

## Evidence
- Failing runs: https://github.com/mixpeek/amux/actions/runs/33900769873 and https://github.com/mixpeek/amux/actions/runs/33901296442 (both: same 2 tests, all 3 browser projects, "6 failed" / "295 passed")
- `error-context.md` from the CI artifact shows the "This worker has no directory to browse" toast, confirming the seeding-never-took-effect diagnosis directly (not inferred)

### DIFF
--- a/e2e/worker-action-parity.spec.ts
+++ b/e2e/worker-action-parity.spec.ts
@@ -26,11 +26,27 @@ const SAMPLE = {
 async function boot(page: import('@playwright/test').Page) {
   await page.goto('/');
   await page.waitForFunction(() => typeof (window as any)._renderWorkerActionMenu === 'function');
-  await page.evaluate(({ sample, name, root }) => {
-    // These are top-level lexical bindings in the classic app bundle, not
-    // window properties. Seed the exact session/root shared by both handlers.
-    eval(`sessions = [${JSON.stringify(sample)}]; peekSession = ${JSON.stringify(name)}; peekSessionDir = ${JSON.stringify(root)}`);
-  }, { sample: SAMPLE, name: NAME, root: ROOT });
+  // `sessions`/`peekSession`/`peekSessionDir` are top-level lexical (`let`)
+  // bindings in the classic app bundle, not window properties — the comment
+  // this replaces already knew that much, but drew the wrong conclusion from
+  // it. A nested eval() INSIDE a page.evaluate(fn, arg) callback still runs in
+  // that callback's own isolated scope (Playwright drives it via
+  // Runtime.callFunctionOn): it cannot see or reassign another script's `let`
+  // bindings, so this silently no-op'd every time. `sessions` stayed `[]`,
+  // `_browseWorkerFiles` computed an empty root from it, and every click in
+  // this file just toasted "This worker has no directory to browse" instead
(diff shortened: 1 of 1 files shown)


ivomarino:
@esteininger `e2e` now fully passes (all 3 browser projects) — this took 4 rounds to get right; see the PR description / commit history for the full diagnosis (a genuine app.js scoping quirk with Playwright's page.evaluate, a view-persistence race, a session-poll race, and a DOM-detachment regression from an earlier attempt, each confirmed against real CI runs, not guessed).

The `check` failure here is unrelated to this diff (pure e2e test file) — it's the pre-existing `test-cli-launch-unbound.sh` break on `main` that #193 fixes. #191 and #193 are mutually blocking each other's CI from reading fully green (each fixes the other's failing check) but each is independently correct and tested — merging either first clears the other's CI on its next run.

ivomarino:
Closing — superseded. `main` independently fixed the same underlying bug this PR targets: commit `f20422b8` ("test(dashboard): keep peek reachability atomic", Ethan, 2026-09-04) fixes the exact same DOM-detachment race via a cleaner mechanism (an atomic in-page reachability check instead of my boot()/bootWithPeek() split). Confirmed `main`'s own `rust` workflow has been green for 3 consecutive runs since. No need to merge this on top.

PR #193, opened 2026-09-05
fix(ci): scope the AMUX-3145 specimen match to cmd_start, not the whole file

ivomarino:
CI red on `main` (and every open PR based on recent main, including
mine — #191) since `AF-496` (`cmd_browser`, landed 2026-09-04).

## Root cause

`scripts/test-cli-launch-unbound.sh`'s self-check asserts the
`local AMUX_API="${AMUX_API:-$(cmd_url)}"` specimen line is unique in
the whole `amux` file, then builds a "broken" copy by deleting it and
verifies the launch smoke fails on that copy.

`cmd_browser` (AF-496) legitimately needs the identical resolution for
the identical reason and copied the line verbatim — a second correct
use, not a second instance of AMUX-3145. The self-check's whole-file
uniqueness assumption just went stale the moment any other function
needed the same pattern:

```
FAIL could not build the broken fixture; has the specimen line drifted?
     expected exactly 1 occurrence in ./amux and 0 in the copy; got 2 and 0
```

## Fix

Match/delete the specimen within `cmd_start`'s own function body
(bounded by its header line and the next column-0 `}`) instead of the
whole file — the thing that needs to stay unique is `cmd_start`'s own
declaration, not the literal string globally.

Verified locally against the real `./amux`: confirmed the unpatched
script fails exactly as CI shows (1 pass / 2 fail), then all 3 checks
pass after this fix.

## Evidence
- Failing on `main` directly: https://github.com/mixpeek/amux/actions/runs/33911178826
- Also failing on #191 (unrelated to that PR's own diff): https://github.com/mixpeek/amux/actions/runs/33910929803

### DIFF
--- a/scripts/test-cli-launch-unbound.sh
+++ b/scripts/test-cli-launch-unbound.sh
@@ -35,10 +35,25 @@ PASS=0; FAIL=0
 ok()  { echo "  ok   $1"; PASS=$((PASS+1)); }
 bad() { echo "  FAIL $1"; [ -n "${2:-}" ] && echo "       $2"; FAIL=$((FAIL+1)); }
 
-# The one line whose removal reproduces AMUX-3145. Unique in the file: it is the only
-# AMUX_API declaration resolved via $(cmd_url) (the others use a literal default), so
-# deleting it leaves the use at amux:650 out of scope without touching any other path.
+# The one line whose removal reproduces AMUX-3145: the only AMUX_API declaration
+# resolved via $(cmd_url) (the others use a literal default), so deleting it leaves
+# the use at amux:650 out of scope without touching any other path.
+#
+# NOT necessarily unique in the whole FILE, though — AF-496 (2026-09-04) added
+# cmd_browser, which legitimately needs the exact same resolution for the exact same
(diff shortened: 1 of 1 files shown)


ivomarino:
@esteininger this fixes a real CI break on `main` (the `test-cli-launch-unbound.sh` self-check went stale when AF-496's `cmd_browser` legitimately reused the same specimen line it assumed was file-unique — see PR description for the full root cause). `check` job here passes.

The `e2e` failure showing on this PR is unrelated to this diff (a pure bash test-script change) — it's the still-open `worker-action-parity.spec.ts` bug that #191 fixes. #191 and this PR are mutually blocking each other's CI from reading fully green (each fixes the OTHER's failing check), but each is independently correct and tested. Merging either first will make the other's CI clear on its next run.

ivomarino:
Closing — superseded. `main` independently fixed the same underlying bug this PR targets: commit `3d77f4d1` ("test(cli): scope launch negative control", Ethan, 2026-09-04) fixes the same `test-cli-launch-unbound.sh` staleness this PR addresses. Confirmed `main`'s own `rust` workflow has been green for 3 consecutive runs since. No need to merge this on top.

PR #194, opened 2026-09-05
fix(email-themes): call registry::tick() so the loop can report its own health (AMUX-123)

ivomarino:
`email-themes` was showing `stalled` (`ticks:0`, `last_tick_at: null`) for
its entire ~17h uptime — not evidence the loop is dead, evidence it never
called anything the registry's `classify()` reads.

## Root cause

`theme_refresh_loop()` is the "do the work, then sleep for the interval"
shape every sibling loop uses (`pipe_reconcile_loop`, etc.), and every one
of those calls `registry::tick(id)` once per iteration. This one never
did — a pure instrumentation gap, not a real stall: `classify()`'s own
"never ticked past the stall budget" rule reports `stalled` for a loop
like this permanently, on every run, whether or not real work happened.

## Fix

Call `registry::tick(ids::EMAIL_THEMES)` once per loop iteration, after
`one_theme_pass()` — including its "skip, themes still fresh" branch,
since the loop's job is deciding whether to recompute, not only the
recompute itself, so a skip is a real healthy pass.

Two-fix rule: the fix itself is the log signal — `classify()` already
surfaces ticks/`last_tick_at` at `GET /api/system-jobs`, so once this loop
actually reports, a real future stall becomes visible the same way every
other job's does.

Verified locally: `cargo check --workspace --all-targets` and `cargo
clippy --workspace --all-targets -- -D warnings` both pass (via the
repo's own commit hook).

### DIFF
--- a/crates/amux-server/src/api/email_intel.rs
+++ b/crates/amux-server/src/api/email_intel.rs
@@ -689,6 +689,18 @@ pub async fn theme_refresh_loop() {
     // and the mechanism disagree.
     loop {
         one_theme_pass().await;
+        // AMUX-123: this loop never called the registry's tick() — every
+        // sibling loop of this exact "do the work, then sleep for the
+        // interval" shape does (see pipe_reconcile_loop). Without it, `ticks`
+        // stays 0 and `last_tick_at` stays null FOREVER, regardless of
+        // whether the loop is actually healthy: classify()'s own "never
+        // ticked past the stall budget" rule then reports this job as
+        // `stalled` on every single run, permanently, whether or not a real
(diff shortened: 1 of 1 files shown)


ivomarino:
@esteininger this fixes a real instrumentation gap (`email-themes`'s loop never called `registry::tick()`, so it permanently misreported as `stalled` regardless of whether it actually ran — see PR description). `e2e` and `checks` both pass; verified locally with `cargo check --workspace --all-targets` + `cargo clippy --workspace --all-targets -- -D warnings`, both clean.

The `check` failure here is unrelated to this diff — it's the pre-existing `test-cli-launch-unbound.sh` break on `main` that #193 fixes. Will clear once #193 merges.

PR #195, opened 2026-09-05
feat(build): amux owns amux-rust-base freshness on the Woodpecker fleet (AMUX-125) + worker-start honesty fix (AMUX-120)

ivomarino:
Two independent, already-verified fixes that hadn't gone upstream yet.

## AMUX-125: rebuild-rust-base.sh

infra had been rebuilding this repo's own `amux-rust-base` image (built
from `Dockerfile.rust-base`) on all 3 Woodpecker build hosts via a
`docker_image` Tofu resource in each of its own host-provisioning modules
— a side effect of provisioning, not something infra should keep owning.
Ivo's call, relayed via infra: amux owns this image's freshness now.

`scripts/rebuild-rust-base.sh` fans `docker build` out to every host in
parallel via pre-created `docker context`s (same shape as the existing
`rust-remote-build.sh`), reading context names from
`AMUX_RUST_BASE_CONTEXTS` so no hostnames/IPs land in this public repo. A
slow/failed host doesn't block the others. Plus doc/comment updates in
`.woodpecker.yml`, `Dockerfile.rust-base`, `docs/woodpecker-setup.md`
pointing at it (and correcting a couple of details that went stale within
the hour they were written — a host rename and a transport change).

**Not yet operational**: infra's `tofu/docker-mtls/` module is the only
transport into these hosts' dockerd now (TCP + mutual TLS on 2376,
replacing SSH fleet-wide) and this box doesn't have the client cert yet —
asked infra for it. infra's existing 3 Tofu resources stay in place,
deliberately, until this script has been verified working against all
three hosts at least once.

## AMUX-120: worker-start.sh honesty fix

`amux-worker-start.service`'s boot script counted a lane as "started" as
soon as the API accepted the launch request, never verifying a claude
process actually landed — found live when a `~/.bashrc` boot-time PATH
break left 5 of 7 lanes dead for 4+ hours while the script's own log
claimed "7 started, 0 failed" the whole time. Now polls the session's real
`running` state for ~10s before counting it as started.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #196, opened 2026-09-06
fix(ci): the cmd_start window never closed on the runner's awk, so `check` was red for everyone

esteininger:
## What

`scripts/test-cli-launch-unbound.sh` scopes its specimen count to `cmd_start()` with an awk window closed by `/^}/`. An unescaped `}` outside an interval is **undefined in POSIX ERE**. BSD awk (macOS) reads it as a literal and closes the window; the CI runner's awk does not, so `in_cmd_start` never resets and the count runs to EOF.

## Why it went red on 2026-09-04 and not before

Dormant until `daf17cd5` (AF-496) added a second, entirely legitimate `local AMUX_API="${AMUX_API:-$(cmd_url)}"` inside `cmd_browser`. From that commit on, every PR whose merge base included it fails:

```
FAIL could not build the broken fixture; has the specimen line drifted?
     expected exactly 1 occurrence in ./amux and 0 in the copy; got 2 and 0
```

The `2` is the proof. Reaching `cmd_browser` at line 4018 from `cmd_start` at line 636 requires the window to have stayed open across ~3400 lines and every function boundary between them.

## Cost

PR #195 carried this as its own red for two days and never touched `amux`. On macOS the suite passes 3/3, so reproducing it needs a Linux awk — the exact shape that gets written off as a flake.

## Changes

- Both windows (the counter and the fixture builder) use `substr($0,1,1) == "}"`. No regex, so the two platforms cannot disagree.
- The failure text no longer asserts one cause. When the in-window count equals the file-wide count it says the window never closed and names the awk. "has the specimen line drifted?" sent the last reader into `./amux`, where nothing was wrong.
- A file-wide counter, used only to tell those two causes apart (ethos rule 4: name what should appear *beside* the answer).

## Verified

```
bash scripts/test-cli-launch-unbound.sh   -> 3 passed, 0 failed
in-window 1, file-wide 2                  (the discriminator now has both numbers)
```

The real proof is this PR's own `check` job going green, since that is the platform the bug lives on.

### DIFF
--- a/scripts/test-cli-launch-unbound.sh
+++ b/scripts/test-cli-launch-unbound.sh
@@ -42,13 +42,27 @@ bad() { echo "  FAIL $1"; [ -n "${2:-}" ] && echo "       $2"; FAIL=$((FAIL+1));
 # touching another function's independent declaration.
 SPECIMEN='local AMUX_API="${AMUX_API:-$(cmd_url)}"'
 
+# `substr($0,1,1) == "}"`, NOT /^}/. An unescaped `}` outside an interval is
+# undefined in POSIX ERE: BSD awk (macOS) reads it as a literal and closes the
+# window, the CI runner's awk does not, so `in_cmd_start` never resets and the
+# count runs to EOF. That is not hypothetical -- it is what turned this gate red
+# for every PR from 2026-09-04, when daf17cd5 (AF-496) added a SECOND legitimate
+# `local AMUX_API=...` inside cmd_browser. The window silently swallowed it, the
+# count came back 2 instead of 1, and the failure text below blamed the specimen.
+# Written this way so the two platforms cannot disagree at all.
 count_cmd_start_specimen() {
   awk -v needle="$SPECIMEN" '
     /^cmd_start\(\)[[:space:]]*\{/ { in_cmd_start=1 }
     in_cmd_start && index($0, needle) { count++ }
-    in_cmd_start && /^}/ { in_cmd_start=0 }
+    in_cmd_start && substr($0, 1, 1) == "}" { in_cmd_start=0 }
     END { print count + 0 }
   ' "$1"
+}
+
+# File-wide count, no window. Only used to tell the two failure causes apart in
+# the diagnostic below (ethos rule 4: name what should appear BESIDE the answer).
+count_file_wide_specimen() {
+  awk -v needle="$SPECIMEN" 'index($0, needle) { count++ } END { print count + 0 }' "$1"
 }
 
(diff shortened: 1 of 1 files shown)


esteininger:
Daily review verdict: **leave open until the runner proves the portability fix.**

Evidence read: the complete current 64-line diff in the sole file `scripts/test-cli-launch-unbound.sh` at `e2bbc54b`; GitHub reports MERGEABLE/BLOCKED, and `gh pr checks 196` currently has `check`, `checks`, and `e2e` all pending.

The change addresses the stated root cause: both awk windows replace the undefined POSIX ERE `/^}/` with the character comparison `substr($0, 1, 1) == "}"`. The added whole-file count also makes a future never-closed window distinguishable from a genuinely drifted specimen, so the failure would surface with the evidence needed to diagnose it. The negative fixture still exercises the actual broken CLI path.

What its local 3/3 result cannot prove is the bug's defining property: behavior on the Linux runner awk that disagreed with BSD awk. The PR itself correctly names its `check` job as the real proof; that job is not complete yet. This is a test-only path, so there is no fleet-wide launch/runtime merge restriction once CI is green, but green is still necessary.

Owner/next action: @esteininger to wait for the existing CI run; merge only if the Linux `check` and the remaining required jobs pass. No rerun/poll loop is warranted.

esteininger:
Closing this — the diagnosis was wrong and the fix is unnecessary.

I read #195's `check` failure as the runner's awk failing to close the `/^}/` window. It isn't. #188 and #160 both ran step 10 ("runtime smoke -- CLI launch unbound-var") on the same runner today and **passed**, so the window closes fine there.

The real cause is in the failure text I quoted and didn't read closely enough:

```
expected exactly 1 occurrence in ./amux and 0 in the copy; got 2 and 0
```

No `cmd_start` in that sentence. `3d77f4d1` (ATE-44, 2026-09-04 19:45) added that word when it scoped the count to `cmd_start()`. So #195's job executed the **pre-`3d77f4d1`** script, which counted file-wide — and file-wide, `amux` has had 2 occurrences since `daf17cd5` added a legitimate second one in `cmd_browser`. On that version, 2 is the correct answer and the failure is real.

#195's merge commit predates the script fix, and `gh run rerun` re-runs the same commit rather than recomputing the merge, so the re-run reproduced it exactly. Only a new push refreshes it. The remedy is a rebase of #195, not a change here.

`/^}/` is still undefined ERE in principle, but both platforms in use close the window correctly, so this would be a fix pinning a layer that is not broken — and shipping it under a commit message asserting a false cause is worse than not shipping it.

PR #197, opened 2026-09-07
fix(telegram): surface messages_group_ignored on GET /api/telegram/status

ivomarino:
The `Report` struct already tracked this counter — its own doc comment says it exists so a group-ignored pattern is sweep-catchable, not just visible via `tracing::warn` — but the status handler never included it in the JSON response.

Found while looking into linking a Telegram group for real use: this field is exactly what will show whether messages posted in a linked group are landing (routed via `@lane` mention) or being silently ignored (no mention, by design) — previously only visible by grepping raw logs.

One-line fix, no behavior change to the poll loop itself.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -12,7 +12,9 @@
 # pipeline does NOT build it itself: baking the toolchain + a warm dependency
 # cache into every run is the exact slowness this pipeline exists to avoid.
 # Rebuild it manually whenever Cargo.lock changes enough that the cache goes
-# meaningfully stale.
+# meaningfully stale — scripts/rebuild-rust-base.sh does this across all
+# real build hosts at once (amux's own job as of 2026-09-05, see
+# docs/woodpecker-setup.md).
 #
 # Structural advantages over the ad-hoc `docker build` approach it replaces,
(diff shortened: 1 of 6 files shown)


esteininger:
Reviewed all six files in `gh pr diff 197` at dd8cacc5c10fc0349cf320ff699dae7437fa11b3 and the current `gh pr checks 197` results. **Leave open; not ready to merge.**

The effective scope is smaller than the historical diff: against clean main 7321e562, all five infrastructure/startup files are byte-identical; only the added `messages_group_ignored` status field remains. I am not treating already-landed worker-start changes as new work in this PR.

The field correctly exposes the existing runtime counter, but no added endpoint assertion would catch its omission or a wrong source field. Please add a focused status-response regression test and refresh CI against main. The failing E2E job [101690930537](https://github.com/mixpeek/amux/actions/runs/34105969568/job/101690930537) reports `working-now-accuracy.spec.ts:40` on iOS Safari: expected a current doing-card id, received an empty string (1 failed / 309 passed / 32 skipped). I have not established whether that is a branch regression or shared fixture behavior, so I will not dismiss it as a flake.

Owner: @ivomarino for the counter test and refreshed branch; maintainer review after green CI and examination of that test's failure coverage.

PR #216, opened 2026-09-11
feat(provider): Muse Code worker provider

reluctantfuturist:
Adds Muse Code as an amux worker provider, alongside claude/codex/gemini/ollama. Run in production on a real fleet for the last several days — the later commits are defects that only showed up there, kept separate so the reasoning survives.

## What it adds

- `muse` in `SESSION_PROVIDERS`, the provider adapter, the model catalog and the create modal
- Launch, resume and send, with muse's own session model (below)
- Hooks: muse delivers Claude-pattern hooks via plugin, so `MuseAdapter` is the only non-Claude adapter with `hooks: true`. `docs/provider-parity.md` scores it and rows 2 and 11 move to MET
- An e2e spec that is meaningful in CI without muse installed (create modal, stored provider/flags, worker card) and does a live launch when `muse` and `tmux` are both on PATH

## The part worth reviewing closely

**Muse cannot be told its session id.** There is no `--session-id` to mint, so a first start launches bare and amux learns the id from disk afterwards. Doing that naively is a data-loss bug that hides perfectly: if discovery silently fails, everything still looks fine and the lane simply opens a brand-new conversation on every restart.

So the id is taken from a **before/after diff** of muse's session directory, snapshotted before the process exists — scanning only afterwards cannot tell a session this lane just opened from one another lane opened a second earlier. `muse_pick_session` refuses to guess when two appear in one workspace, and the e2e test reads the meta the server wrote and requires a real uuid rather than trusting the launch to have worked.

An earlier version keyed discovery on `workspace_root`, which muse writes lazily — it was wrong in exactly the case it existed for.

## Defects found in production, each its own commit

- **`amux start` bypassed the server** for muse, so the lane launched without the server's own launch path. Routed through the server; provider-parity scored.
- **Muse workers did not actually work**: the workspace-trust prompt, a 20ms post-paste settle tuned for Claude (muse needs ~350ms), and Claude-shaped send verification stacked into a lane that looked alive and did nothing. Verification now reads muse's `user_intent.accepted` transcript records.
- **`claude_ui_visible` knew claude and codex markers and none for muse**, so `send_after_ready` polled its whole 60s timeout and DROPPED the start prompt — measured twice in one minute on a live lane, while logging "Claude UI never became ready" for a lane running no Claude, which reads as a launch bug and is not one. Renamed `agent_ui_visible` (nothing about the old name invites you to add a provider to it — that is how muse was missed), muse recognised by two **independent** markers, and the log line now names the provider it watched.
- **The board CLI needed a temp file muse cannot write.** Muse sandboxes writes to the workspace and has no per-directory grant — no `--add-dir`, only `--disable-sandbox` or a named permission profile — so `mktemp` in `$TMPDIR` fails with `Operation not permitted` and `amux board ls` died on a *diagnostic* header capture. A worker could see its work and could not close its card. `_amux_tmpfile` falls back to the working directory and the board path tolerates getting nothing back.

## Testing

12 muse unit tests plus provider conformance; the discovery refusals (two in one workspace, none attributable, other workspaces, trailing slash) are each pinned separately because each was a real wrong answer at some point. The composer test proves each muse marker works **alone** — with both in one sample the suite stays green after either is deleted.

**Note on the suite:** `cargo test -p amux-server` on this branch fails 10 tests. Those same 10 fail on `main` at `065022c1` — identical names, verified by running both — so this branch introduces no regressions. They are in `composer_state_tests`, `submission_gate_tests`, `gate_agreement_tests` and `sessions_legacy::status_truth`, and they look unrelated to this change, but you may want to know main is currently red.

Clippy is clean with `-D warnings`. `APP_VER` and the sw.js `CACHE` are bumped together to 0.9.897.

## Known gap, deliberately not addressed here

`backend/adapter.rs` carries a second, narrower copy of the readiness predicate whose docstring scopes it to Claude panes; it feeds `at_shell_prompt` on the herdr backend only. The tmux path routes through the renamed shared predicate and is covered. Adding muse markers to the herdr copy felt like scope creep for a provider PR, but a muse pane on that backend would still read as a dead shell.


esteininger (commented):
Approved and kicked off the two pending workflow runs, they were sitting on the fork-approval gate rather than actually failing.

This is a strong provider PR. It matches the existing ProviderAdapter shape exactly (registered in default_registry(), SESSION_PROVIDERS, launch_base_binary, provider_label, and the JS-side mirrors), and the two generic fixes underneath it are the right shape: renaming claude_ui_visible to agent_ui_visible and adding a muse sibling to jsonl_user_msg_since are root-cause fixes rather than muse-specific special cases, and both self-announce with a log line naming the provider.

Ran it myself since CI hadn't executed yet: checked out the PR head in a scratch worktree, `cargo check -p amux-server` and `cargo clippy -p amux-server --all-targets -- -D warnings` both came back clean. Session-id discovery (before/after directory diff, refuses to guess on ambiguity, tie-breaks on workspace) has dedicated tests for each failure mode, which is exactly the part of a provider integration that tends to be guessed at instead.

One thing worth a follow-up rather than a blocker: the PR body already flags that backend/adapter.rs's herdr-backend readiness predicate doesn't recognize muse markers yet, so a muse pane on that backend reads as dead. Since that's a known, scoped-out gap rather than an oversight, could you file an issue for it before or right after merge so it doesn't quietly become undocumented?

Let's get the CI runs green and this should be ready.

esteininger:
The checks failure just now is unrelated to this PR, it's test-harness-wired.sh catching a gap I introduced on main an hour ago (a test script I added without wiring into CI). Fixed on main in 7099a099. GitHub Actions pins the merge-base snapshot at trigger time so a plain re-run doesn't pick up the fix; this will go green on the next push/rebase, or I can force it if you'd rather not touch your branch.

esteininger:
Closing and reopening immediately to force a fresh CI check against current main.

esteininger:
The check/e2e failures are the pre-existing main-branch flakiness you already flagged in the PR description (10 failures, all in api::session_verbs/api::sessions_legacy::status_truth, all active-vs-idle detection, none touching muse/provider code). Confirmed unrelated to this PR's content. This machine is also under real load/disk pressure right now from an unrelated process, which is consistent with these being environmental rather than a real regression, but I can't safely re-run cargo here to fully confirm that until the pressure clears. Will recheck once it does; nothing you need to do on your end.

esteininger:
Pushed 89c73ec4: merged origin/main (as of 1fe8c9a4), resolved the APP_VER/CACHE version-bump conflict, and fixed a real compile break the merge exposed -- 3 test call sites in session_verbs.rs still referenced claude_ui_visible after this PR's own rename to agent_ui_visible (a genuine muse-detection fix elsewhere in the same file). cargo check -p amux-server --lib and cargo clippy both pass locally.

Also worth noting: the CI failures on the previous run (10 tests in status_truth/composer_state_tests) were a fleet-wide regression on main itself, unrelated to this PR -- already fixed and merged separately.

main has moved 9 more commits since my merge base (mostly composer/mobile work touching app.js again), so this may show conflicting once more until I re-sync. Will follow up.

